PR #193, opened 2026-06-12
feat(engine): bounded-durable mailbox, location routing, §5 cleanup (fleet Phase 2)

willwashburn:
## Summary

Implements Fleet Delivery Phase 2 for AgentWorkforce/relay#1056 on top of `feat/fleet-nodes-engine` / relaycast#192.

- Evolves durable deliveries into the bounded mailbox state machine: `queued -> delivered -> acked`, per-agent monotonic `seq`, cumulative node `delivery.ack`, TTL dead-lettering, and depth-cap reject-new sender feedback.
- Routes inbound delivery by registered location: self-connected agents keep the existing WS push path, while `via_node` agents receive `deliver {agent,msg_id,seq,mode,payload}` over the node control connection.
- Marks cumulative delivery ack as read state so `/v1/inbox` does not resurface acked messages, and preserves unacked deliveries for reconnect/broker-death redelivery.
- Adds mailbox defaults in one config point: TTL 1h, depth cap 1000, with workspace-level overrides through engine config.
- Updates delivery wire types, SDK tests/comments, OpenAPI delivery/node/trigger coverage, and the SQLite migration/schema.

## Stack / Merge Order

This PR is stacked on relaycast#192 (`feat/fleet-nodes-engine`) and should merge after that PR lands. The base requested here is `main`, so the diff includes the stack context until #192 is merged.

## §11 Audit / §5 Cleanup Gate

Audit commands searched the repo for `relay://`, `/v1/ws`, resource subscriptions, and per-agent stream consumers.

Findings:

- No engine-owned `relay://` resource definitions were found to delete in this repo.
- Active consumers still exist outside the engine package: `packages/mcp/src/resources/definitions.ts`, `packages/mcp/src/resources/ws-bridge.ts`, `packages/mcp/src/resources/subscriptions.ts`, and MCP resource/subscription tests still register/read/subscribe `relay://...` resources.
- Per-agent `/v1/ws` is still consumed by the TypeScript SDK (`AgentClient.connect` / subscription helpers), SDK docs/examples, and other SDK surfaces.

Result: the PTY-agent resource/subscription surface is kept in this PR. Actual deletes remain gated behind that audit because consumers still exist.

## Verification

- `npm run typecheck --workspace=@relaycast/engine`
- `npx vitest run src/__tests__/conformance/delivery.test.ts src/__tests__/conformance/node.test.ts src/__tests__/atomicity.test.ts` from `packages/engine`
- `npm run test --workspace=@relaycast/types`
- `npm run test --workspace=@relaycast/engine`
- `npm run test --workspace=@relaycast/sdk`
- `npm run build`
- `npm run test`

Root `package.json` does not define a repo-level `typecheck` script; engine workspace typecheck is the explicit typecheck run.


willwashburn (commented):
Verified `npm run build` and `npm run test` pass after installing workspace deps.

1. major packages/engine/src/routes/deliveryRouting.ts:46 — routing for `via_node` deliveries is based on the `location_type`/`location_node_id` snapshot stored on the delivery row, not the recipient’s current location. If an agent re-registers on a different transport before fanout runs, the message still goes to the stale node/socket and can be lost. Re-resolve the current agent location at send time, or migrate queued deliveries when location changes.

2. major packages/engine/src/adapters/node/realtime.ts:311 — `attachNodeSocket()` adds sockets to a `Set` and never replaces an existing control connection. A reconnect can therefore leave two live node sockets for the same node, duplicating `deliver`/`action.invoke` traffic and any acks. Enforce a single active node socket by closing the prior one before attaching the new connection.

willwashburn:
Addressed the fanout race in : delivery fanout now re-resolves the recipient's live location from  before push, and the mailbox suite now has a regression test that flips the recipient from  to  before dispatch to prove the delivery still lands.

I left the node socket supersede/close change to Fleet1 as requested and will rebase on that tip when it lands.

willwashburn:
Addressed the fanout race in deliveryRouting.ts: delivery fanout now re-resolves the recipient's live location from the agents table before push, and the mailbox suite now has a regression test that flips the recipient from via_node to self_connected before dispatch to prove the delivery still lands.

I left the node socket supersede and close change to Fleet1 as requested and will rebase on that tip when it lands.

willwashburn (commented):
Round 2 review against feat/fleet-nodes-engine. Verdict: NO-GO. GitHub would not allow this authenticated reviewer to request changes on its own pull request, so I am posting the blocking review as a comment.

1. [major] TTL expiry can overwrite a concurrent ack and emit a false delivery_failed. `expireDueDeliveries` selects active rows at packages/engine/src/engine/delivery.ts:398-411, then updates the captured ids at packages/engine/src/engine/delivery.ts:416-425 without re-checking that status is still queued/delivered. If a node `delivery.ack` lands between those awaits, `ackRows` can set the row to `acked`, but the expiry update will still change it to `dead_lettered` and notify the sender. This violates queued -> delivered -> acked terminal success and the round-2 concurrent ack/TTL edge. Make the expiry update status-guarded and build notices only from rows actually transitioned, ideally in one transaction/returning update.

2. [major] Reconnect redelivery does not preserve the live deliver payload and reports the recipient as the message author. Live fanout sends `payload: { type, data }` from `routeDeliveryOutcomes` at packages/engine/src/routes/deliveryRouting.ts:76-85. Reconnect replay rebuilds a different `payload` in `deliverPendingToNode` at packages/engine/src/engine/delivery.ts:462-485 and sends it at packages/engine/src/engine/delivery.ts:523-531. That reconstructed payload also sets `message.agent_id` from `row.delivery.agentId` at packages/engine/src/engine/delivery.ts:477-480, which is the target agent, not the sender. After broker death, an unacked message with the same `msg_id`/`seq` can therefore be reinjected with a different shape and wrong author id. Persist or reconstruct one canonical deliver payload for both live and replay paths, using the original sender id/name, and add a reconnect test that asserts payload equality, not only `msg_id` and `seq`.

Tests run:
- `npm ci`
- `npm run build`
- `npm run typecheck --workspace=@relaycast/engine`
- `npx vitest run src/__tests__/conformance/delivery.test.ts src/__tests__/conformance/node.test.ts src/__tests__/atomicity.test.ts` from `packages/engine`
- `npm run test --workspace=@relaycast/types`
- `npm run test --workspace=@relaycast/sdk`
- `npm run test --workspace=@relaycast/engine`
- `npm run test`

willwashburn:
Round 2 review items addressed in ea89632:

- TTL expiry now guards active status in the UPDATE itself, and expired-notice fanout only uses rows actually updated, so an acked row cannot be dead-lettered or falsely notify the sender.
- Reconnect redelivery now uses the same shared deliver frame builder as live fanout, and the replay path reconstructs the same payload shape from the persisted message row.
- Added concurrent ack-vs-TTL and live-vs-replay payload-equality regressions.

Fleet1-owned single-socket-per-node work remains untouched here as requested.

willwashburn (commented):
Round 3 review: NO-GO. GitHub rejected `--request-changes` for this account with: `Review Can not request changes on your own pull request`, so this is posted as a review comment with the same blocking findings.

1. [blocker] PR 193 is not rebased on the PR 192 head fix and reintroduces multi-socket node delivery. `git merge-base --is-ancestor origin/feat/fleet-nodes-engine HEAD` is false for `ea89632` vs `8949f842`, and the current code stores `nodeSockets` as a `Set` at `packages/engine/src/adapters/node/realtime.ts:66`, then sends every control frame to every socket at `packages/engine/src/adapters/node/realtime.ts:244`, while `attachNodeSocket` only adds the new socket at `packages/engine/src/adapters/node/realtime.ts:311`. A reconnect with an old socket still present can duplicate `deliver` and `action.invoke` frames, which is exactly the single-socket class of bug round 3 asked us to verify. Rebase or merge `feat/fleet-nodes-engine` tip, keep one current socket per node, close the superseded socket, and keep the reconnect/no-duplicate test on this stack.

2. [major] Broker-death replay does not preserve the live delivery wire frame for non-channel messages. Live DM, group DM, and thread deliveries call `routeDeliveryOutcomes` with `dm.received`, `group_dm.received`, and `thread.reply` at `packages/engine/src/routes/dm.ts:131`, `packages/engine/src/routes/groupDm.ts:185`, and `packages/engine/src/routes/thread.ts:126`, but `deliverPendingToNode` always rebuilds the pending row as `message.created` at `packages/engine/src/engine/delivery.ts:562`. After reconnect, the same delivery can change `payload.type` and data shape, breaking consumers and fixture parity. Persist the original delivery event type/payload or reconstruct through the same builders used by the live fanout path, and add reconnect redelivery tests for DM, group DM, and thread reply deliveries.

3. [major] A future cumulative ack can permanently suppress later replay. `ackRows` advances `agents.deliveryAckSeq` directly to any `up_to_seq` at `packages/engine/src/engine/delivery.ts:313`, while new delivery seqs are assigned from only `MAX(deliveries.seq)+1` at `packages/engine/src/engine/deliveryWrites.ts:66`, and replay filters rows with `deliveries.seq > agents.deliveryAckSeq` at `packages/engine/src/engine/delivery.ts:550`. I reproduced this by sending `delivery.ack {up_to_seq:100}` before the first message: the live frame used `seq:1`, then after node reconnect/inventory sync `replayDeliveries` was `0`. Do not advance the ack cursor past rows that actually exist and were acked, or make seq allocation account for the ack cursor; add a regression for future/stale acks before first delivery.

4. [major] The new TTL-vs-ack race coverage is flaky under fresh execution. A focused run of `npx vitest run src/__tests__/conformance/delivery.test.ts src/__tests__/conformance/node.test.ts src/__tests__/atomicity.test.ts` failed at `packages/engine/src/__tests__/conformance/delivery.test.ts:706` because `expireDueDeliveries` returned one notice in `does not dead-letter an acked delivery when ack and TTL expiry race`. A later forced full `npm run test -- --force` passed, which points to nondeterministic scheduling in the test or implementation. Make this deterministic: either enforce the intended ack-wins ordering in storage, or adjust the test to assert exactly one terminal winner and no duplicate sender fanout.

Verification run from `/tmp/review-Rev2x3`:
- `npm ci`
- `npm run build --workspace=@relaycast/types` and `npm run build --workspace=@relaycast/a2a` to create workspace `dist/` outputs for fresh typecheck
- `npm run typecheck --workspace=@relaycast/engine` passed
- focused engine vitest command above failed once on the TTL race
- `npm run test --workspace=@relaycast/types` passed
- `npm run test --workspace=@relaycast/sdk` passed
- `npm run build` passed
- `npm run test -- --force` passed with 18/18 tasks forced, 0 cached

willwashburn (commented):
Verdict: NO-GO. GitHub would not allow this account to request changes on its own PR, but the finding below should block merge until fixed.

1. [major] openapi.yaml:3815 Duplicate path keys make the OpenAPI document invalid. This PR adds detailed `/nodes`, `/nodes/{name}`, `/triggers`, and `/triggers/{id}` specs at openapi.yaml:2848, but leaves the old stub definitions starting at openapi.yaml:3815. YAML maps cannot contain duplicate keys; running `node -e "const fs=require("fs"); const YAML=require("yaml"); YAML.parse(fs.readFileSync("openapi.yaml","utf8"));"` fails with `Map keys must be unique at line 3815`. Generators/parsers cannot reliably consume the API spec, and less-strict parsers will shadow the detailed node/trigger request schemas/security with the later generic stubs. Remove the old stub path blocks or merge their content into the new detailed definitions so each path appears exactly once.

2. [nit] memory/workspace/.relay/state.json:1 Generated relay sync state changed only timestamps/counters. Drop this file from the PR or ignore generated state so the fleet mailbox change does not carry local runtime metadata.

Verification run in /tmp/review-Rev2x4:
- `git merge-base --is-ancestor 8949f842 HEAD` passed at c39fc7912aeb0e208c8ad318fccf2fcf9c1a88f1.
- `npm install` completed; initial root `npm run build`/`npm run test` passed via Turbo cache.
- Direct runs passed: `npm run build --workspace=@relaycast/engine`, `npm run typecheck --workspace=@relaycast/engine`, `npx vitest run src/__tests__/conformance/delivery.test.ts src/__tests__/conformance/node.test.ts src/__tests__/atomicity.test.ts` from packages/engine, and `npm run test --workspace=@relaycast/types`.
- Round-4 target checks: replay/redelivery now preserves original payload event types via the shared builder and equality regressions; TTL expiry update is status-guarded; fanout re-resolves current recipient location before routing.

PR #194, opened 2026-06-14
feat(engine): per-workspace fleet rollout flag + migration single-delivery guarantees (Phase 6)

willwashburn:
### 1. Per-workspace fleet rollout flag
Gates the **entire fleet node control surface** behind a per-workspace
`fleet_nodes_enabled` flag (default **OFF**) so fleet can ship dark and roll out
workspace-by-workspace. Legacy per-agent WS delivery is unaffected either way.

The flag is checked once at each genuine boundary (no scattered checks):
- **node control WS** (`/v1/node/ws`) → `fleet_nodes_disabled` (404) when off
- **node roster routes** (`/v1/nodes*`) → flat 404 via `requireFleetNodes`
- **declarative trigger evaluation** → skipped at the message hook
- **spawn placement + node-handler dispatch** → refuse in `invokeAction`
  (agent-handler actions stay available)

KV override with a short cache, defaulting to `EngineConfig.fleetNodesEnabled`.
`GET`/`PUT /v1/workspace/fleet-nodes` toggles the per-workspace override. The
`relaycast-engine` serve bin also reads `RELAYCAST_FLEET_NODES_ENABLED` and
`RELAYCAST_MAILBOX_TTL_MS` / `RELAYCAST_MAILBOX_DEPTH_CAP` for self-host tuning.

**Migration single-delivery**: a test proves a legacy self-connected agent is
never *also* delivered via a node when the flag flips mid-stream (one active
location per agent; a node's `agent.register` for an already-connected agent is
rejected `agent_location_conflict`).

### 2. Two cross-repo compat fixes surfaced by the Phase 6 E2E
The engine **runtime deviated from the frozen wire contract (#191)** in two
places despite the golden fixtures, so a real broker could never bring a node
online or complete a spawn. Both are fixed engine-side here (the broker is
contract-correct; **no Rust release**). Neither reopens frozen #192.

- **Node-token transport** (engine read-side #192 ↔ broker send-side #1107): the
  Node HTTP-server `upgrade` handler in `entrypoints/node.ts` read the token only
  from `?token=`, but the broker's `node_control` client sends
  `Authorization: Bearer <nt_live_…>`. It also had **no** fleet-flag gate for
  `/v1/node/ws`. Fixed: read the token from query **or** Bearer header, and gate
  the upgrade behind `isFleetNodesEnabled`.
- **`agent.register` reply frame** (token authority): the broker awaits a canonical
  `reply` frame `{v, id, ok:true, data:{agent_id, token, name?}}` keyed by request
  id (`deny_unknown_fields`); the engine emitted a bare `{type:'agent.registered', …}`,
  so `register_fleet_agent_token` hung to its 30s timeout and **spawn never
  completed**. Fixed: emit the canonical reply frame.

### Regression guards
- `nodeUpgradeAuth.test.ts`: a real WS client authenticates via **both** the Bearer
  header and the query param; flag-off → 404; missing/malformed token → 401.
- The relay two-node E2E (relay#1131) asserts a real broker boots online (Bearer)
  and a spawn **completes** end-to-end (token mint+inject) — these regressions
  cannot reship.

### Version skew
Old v8.x brokers keep working (fleet flag off). Fleet nodes require the new
broker + engine + the per-workspace flag on. The hosted (Cloudflare DO)
equivalent of these fixes is handled in PR 5.

### Tests
Full engine suite green (111). typecheck + targeted eslint clean.


willwashburn (commented):
**RevA1 — adversarial review, round 1.** Built the head branch in a clean worktree: `typecheck` ✅, targeted `eslint` ✅, full engine suite **111/111** ✅ (the 2 new files, 9 tests, execute and pass). Verified the four claimed boundaries are genuinely gated; the migration guarantee and the Bearer/query compat fix both hold under test. No blockers or majors. Findings below are all **minor/nit** — none block merge, but a couple touch the "every node surface inert when OFF" claim.

---

**1. [minor] Token precedence is query-first, but the design calls for header precedence — and the both-present case is untested.**
`engine.ts:501` and `entrypoints/node.ts:113` both do `url.searchParams.get('token') ?? bearer`, i.e. **query wins** when both are present. The Phase-6 brief specifies *header* precedence. In practice the broker sends only the header and SDK/Pear only the query, so no shipping transport sends both — impact is low — but the precedence is a silent, untested contract. `nodeUpgradeAuth.test.ts` only covers header-only and query-only in isolation. Pick a precedence deliberately, document it in the comment, and add a both-present assertion so it can't drift.

**2. [minor] The `invokeAction` fleet gate is fail-OPEN while every other gate is fail-CLOSED.**
`engine/action.ts:357` defaults `const fleetNodesEnabled = options.fleetNodesEnabled ?? true`. The WS gate, `requireFleetNodes`, the trigger hook, and the workspace routes all default `config.fleetNodesEnabled ?? false` (fail-closed). Today there are only two callers — `routes/action.ts` threads the resolved flag, and `engine/trigger.ts:180` relies on the upstream `message.ts` gate (so it passes nothing → defaults true). That works *only* because the message hook gates trigger evaluation. Any future caller that forgets the option silently re-opens spawn + node-handler dispatch with the flag off. Make the option required, or default it `false`, so the placement gate fails closed like the others.

**3. [minor] The declarative-trigger *management* surface is not gated — only evaluation is.**
`routes/trigger.ts:22-76` (`POST/GET/PATCH/DELETE /v1/triggers`) carry only `requireAuth`, no `requireFleetNodes`. The PR body says "flag OFF → every node surface inert," and declarative triggers are explicitly part of the fleet control surface; `fleetRolloutFlag.test.ts:188` even asserts `POST /v1/triggers` → 201 while OFF. So a flag-off workspace can fully create/edit/list triggers (they just never fire). That's a deliberate choice but it contradicts the "every surface" wording and is the second ungated node boundary. Either add `requireFleetNodes` to the trigger CRUD routes for consistency, or document the intentional exception and soften the claim.

**4. [minor] Disabling the flag does not tear down live node connections or re-home `via_node` agents.**
The WS gate guards the *upgrade* (new connections) and `agent.register` only. A node already connected when a workspace flips OFF keeps receiving `deliver` frames until it reconnects — the ON→OFF kill-switch direction is unguarded and untested (the migration test only exercises OFF→ON, which is the right primary direction). Fine for a forward dark-launch flag, but please either document the flag as forward-rollout-only or close live node sockets on disable so "inert when OFF" is true for in-flight nodes too.

**5. [nit] OpenAPI under-specifies `/workspace/fleet-nodes`.** GET/PUT advertise only a generic `SuccessResponse`; the real payload is `{ enabled, default_enabled, override }`, and the PUT body's valid shapes are `{ enabled: boolean }` XOR `{ mode: "inherit" }`. Tighten the schema so generated clients see the real contract. (`openapi.yaml:18-61`)

**6. [nit] Ambiguous PUT body precedence.** `routes/workspace.ts` lets `{ enabled: true, mode: "inherit" }` silently clear the override (mode wins) rather than 400-ing. Low risk; consider rejecting bodies that set both.

---

**Verified correct (not rubber-stamping):**
- `agent_location_conflict` (`engine/node.ts:349`) genuinely fires for an *active self_connected* agent — the `onConflictDoUpdate.setWhere` predicate (`node.ts:341-344`) only permits the update when the row is inactive or already via-node on the same node. The migration test is non-vacuous: a broken guard would relocate `legacy` to `via_node`, flipping *both* `legacySock == 2` and `nodeDeliveries == 0`.
- Bearer header AND `?token=` query each upgrade to 101 when the workspace flag is ON; 404 when OFF; 401 on missing/malformed — exercised against a real booted Node server.
- Flag default OFF confirmed: `DEFAULT_ENABLED = false`, every call site `?? false`.

Round-1 verdict: **GO** on substance — no blockers/majors. Recommend addressing #1–#4 before flag flip.

willwashburn (commented):
## Adversarial review — round 3 (head b673dfb) — **GO**

Built head + ran the engine suite myself: **build green, `tsc --noEmit` clean, eslint clean on all changed files, 111/111 tests pass** (the two new files — `nodeUpgradeAuth.test.ts`, `fleetRolloutFlag.test.ts` — confirmed to actually execute via `vitest list`). All three round-3 items verify:

**(1) Rollout flag — all 4 boundaries gated, default OFF, flip sane ✅**
- WS upgrade: gated in both `engine.ts:204` (hosted/Hono) and `entrypoints/node.ts:133` (self-host). Node roster: `requireFleetNodes` on POST/GET/GET`:name` (`routes/node.ts:22,36,49`). Triggers: `routes/message.ts:158`. Spawn placement + node-handler dispatch: `engine/action.ts:373,397`.
- Default OFF: `DEFAULT_ENABLED=false`; every call site passes `config.fleetNodesEnabled ?? false`. KV override + 10s cache invalidated on `setFleetNodesOverride` (`lib/fleetNodes.ts:794`). Flip-on/flip-off behavior exercised by both new test files. Survived the rebase intact.

**(2) Bearer-OR-query node-WS auth ✅**
- Both upgrade paths read `query('token') ?? bearer`, case-insensitive `^bearer\s+`, trimmed; missing→401, non-`nt_live_`→401, unknown hash→401, flag-off→404. Token is only hashed for lookup, never forwarded. `nodeUpgradeAuth.test.ts` exercises header-only and query-only against a real `ws` client.

**(3) Canonical `agent.register` reply frame ✅ (code correct)**
- `engine/node.ts:519` emits `{v:1, id:requestId(message), type:'reply', ok:true, data:{agent_id, token, name}}`. Matches `FleetReplyMessageSchema` (`.strict()`) exactly, and `data` carries only `{agent_id, token, name}` — `invocation_id`/`session_ref` correctly excluded, so the broker's `deny_unknown_fields` parse of `AgentRegisterReplyData {agent_id, token, name?}` is satisfied. `id` echoes `message.id` when present → id-correlated.

### Non-blocking findings (minor / nit)

1. **[minor] Success-path id-correlation is untested.** `engine/node.ts:520` is the entire point of the fix (broker matches `pending_agent_registrations` by `reply.id`), yet no test asserts a *successful* reply's `id` equals the request `id`. `delivery.test.ts:110` sends `agent.register` with **no** `id` and never reads `reply.id`; only the *conflict/error* path (`fleetRolloutFlag.test.ts:236`, id `claim-legacy`) checks correlation. A regression to `id: generateId()` on the success branch would pass the whole suite — so the PR's "these regressions cannot reship" claim doesn't actually hold for this property. Fix: in `delivery.test.ts` `registerViaNode`, send `id: 'reg-<name>'` and `expect(reply.id).toBe('reg-<name>')`.

2. **[minor/informational] Frozen contract guard eroded (originated in base, not this PR).** `git diff ee2c001 -- packages/types` is **not** empty: `AgentRegisterReplyDataSchema` and the golden `fixtures/fleet-wire/reply.agent_register.json` were **deleted**, and `reply.json`'s `data` was reshaped to `{agent_id,name,token,invocation_id,session_ref}`. These come from base-branch commits (8518e1a, 33ede3c, ddffa76), **not** #194 — #194 touches zero files under `packages/types`, so the "engine-only, no Rust release" claim holds. But the cross-repo parity for the register-reply `data` shape now rests solely on the engine's hand-built object + the broker's Rust struct, with no types-level schema or golden fixture pinning it. Worth restoring a fixture/schema guard in a follow-up so finding #1's correlation + shape can't silently drift again.

3. **[nit] `requireFleetNodes` runs before `rateLimit`** on all three node routes (`routes/node.ts:22,36,49`), so flag-off requests return 404 without passing the rate limiter. Each call is a cached KV read, so impact is negligible, but ordering is inverted vs. the usual `…, rateLimit, handler` convention.

4. **[nit] Precedence is query-over-header, not header-over-query.** `query('token') ?? bearer` means a `?token=` wins over the `Authorization: Bearer` header if both are sent. No practical impact (broker sends header only, SDK query only), but it's the opposite of "header precedence" — pick one and note it.

Verdict: all three round-3 items verified, nothing new at blocker/major. **GO.**

willwashburn (commented):
## Round 4 (adversarial) — GO ✅ clean

Independent third completed pass on head `b673dfb`. Built and tested from a fresh worktree of `origin/feat/fleet-rollout-flag`. No blockers or majors found; confirming rounds 1 & 3.

### Independent verification
- **Typecheck** (`tsc --noEmit`, after building `@relaycast/types`+`@relaycast/a2a`): clean
- **Engine suite** (`vitest run`): **111/111 passing**, 13 files
- **ESLint** on all 11 changed engine files: clean

### Round-4 confirmation criteria
1. **Flag gates all 4 node boundaries, default OFF** ✅
   - WS upgrade: `engine.ts:202` (Hono/in-process) + `entrypoints/node.ts:131` (self-host) → 404 `fleet_nodes_disabled`
   - Roster routes `/v1/nodes*`: `requireFleetNodes` on POST/GET/GET:name (`routes/node.ts:22,36,49`) → flat 404
   - Trigger eval: skipped at the message hook (`routes/message.ts:153-163`)
   - Spawn placement + node-handler dispatch: refuse in `invokeAction` (`engine/action.ts:362,397`); agent-handler actions unaffected (asserted in test)
   - Default OFF: `DEFAULT_ENABLED=false` (`lib/fleetNodes.ts:22`) + `config.fleetNodesEnabled ?? false` at every call site.
2. **/v1/node/ws Bearer-OR-query auth** ✅ — `Authorization: Bearer` OR `?token=` (`engine.ts:185-189`, `entrypoints/node.ts:109-113`); malformed/missing → 401 (`nodeUpgradeAuth.test.ts`); flag-off → 404; `nt_live_` prefix enforced before hash lookup.
3. **Canonical reply frame** ✅ — `agent.register` emits `{v:1, id:requestId(message), type:'reply', ok:true, data:{agent_id, token, name}}` (`engine/node.ts:522`), id-correlated, **not** `agent.registered`. Minimal 3-field `data` matches the broker's `deny_unknown_fields` + #191 `AgentRegisterReplyData`. Conformance test updated to assert the `reply` frame (`delivery.test.ts`). Error path is symmetric: `{v:1, id, type:'error', ok:false, code, message}` matches the strict `FleetErrorMessageSchema`.
4. **Frozen contract** ✅ (with one clarification, below) — PR #194's diff vs its base (`feat/fleet-mailbox`) touches **zero** `packages/types` files; the wire contract is untouched by this PR.

### Minor / informational (non-blocking)
1. **(nit, pre-existing, not #194)** `git diff ee2c001 -- packages/types` is *not* empty — but those changes (`reply.json` gaining `invocation_id`/`session_ref`, deletion of `reply.agent_register.json`) belong to **earlier PRs in the stack**, not #194. The runtime engine emits only the 3-field `data`, so the broker `deny_unknown_fields` contract is satisfied. Note for the stack owner: the `reply.json` fixture's `data` now carries 5 fields while the engine emits 3; harmless because the engine's reply `data` is freeform JSON, but a broker decoding that fixture verbatim with `deny_unknown_fields` would reject the extra fields. Worth a glance in the broker repo's fixture-parity check.
2. **(nit)** `nodeUpgradeAuth.test.ts` `tryUpgrade` resolves only on `open`/`unexpected-response`; a bare socket `error` (no HTTP response) would hang the promise. Fine against a live localhost server, but slightly fragile.
3. **(nit)** `openapi.yaml` documents the GET/PUT `/workspace/fleet-nodes` responses as generic `SuccessResponse` rather than the actual `{enabled, default_enabled, override}` shape — consistent with the existing `/workspace/stream` docs, so cosmetic only.

**Verdict: GO.** Phase 6 rollout flag + the two cross-repo compat fixes are correct and well-guarded.

PR #227, opened 2026-07-01
feat(engine): optional egress proxy for http_push delivery (use_proxy)

khaliqgant:
## What

Adds a per-node opt-in — `delivery.use_proxy: true` — that routes an http_push webhook POST through a deployment-configured forwarder instead of hitting the destination directly.

## Why

Some webhook receivers sit behind **Cloudflare bot protection that blocks the engine's own network origin** — a Cloudflare Worker gets a zone-level **403** on every request (verified exhaustively: bare GET to the receiver's homepage from a Worker → 403; identical request from curl / AWS / GCP → normal). So for those receivers the webhook can *never* be delivered directly from the Worker. Routing through a **non-Cloudflare proxy** gets through.

This generalizes the one-off proxy we stood up for a specific agent into a first-class, reusable flag any http_push node can opt into.

## How

- `httpDeliverySchema` gains `use_proxy?: boolean` (stored in the node's `deliveryConfig`; `url` stays the **real** destination).
- `EngineConfig.httpPushProxy?: { url, secret }` — the deployment's single forwarder, wired by the host adapter from env/secrets.
- `dispatchHttpPush`: when `use_proxy` is set, POST to the proxy URL with:
  - `X-Forward-To: <real destination>`
  - `X-Proxy-Auth: <shared secret>`
  
  The real `url` is still SSRF-checked; the proxy URL is operator-configured and trusted. If `use_proxy` is set but **no proxy is configured**, the delivery fails with a clear error (`no http_push proxy configured`) rather than silently going direct (which would just be blocked).

The forwarder (non-CF host) validates `X-Proxy-Auth`, reads `X-Forward-To`, and forwards the body + webhook auth headers.

## Tests (19/19 conformance pass)
- Routes through the proxy with the right control headers and preserves the node's own webhook auth header.
- Fails cleanly when `use_proxy` is set without a configured proxy (delivery stays queued, error recorded, no webhook POST).

## Deploy to activate (host side)
1. Publish this as the next `@relaycast/engine`.
2. In `relaycast-cloud`: bump the engine, set `HTTP_PUSH_PROXY_URL` / `HTTP_PUSH_PROXY_SECRET`, wire them into `EngineConfig.httpPushProxy` (adapter change staged), redeploy.
3. Register the target node with `delivery.use_proxy: true` and the real `url`.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khaliqgant:
Thanks all — addressed the review in `6399c6c` (rebased on top of the reviewer-bot fixes in `2a59766`). Summary:

**SSRF / header-injection (Gemini HIGH, cubic P1)** — Fixed at the root: added `x-forward-to` and `x-proxy-auth` to `RELAYCAST_CONTROLLED_HEADERS`, so operator `static_headers` can never set them (the check is case-insensitive via `name.toLowerCase()`). This covers *both* the durable and ephemeral paths, so a case-variant `x-forward-to` can't survive into `buildHttpPushHeaders` output and can't override the engine-set control header. Added a test that registers a malicious `x-forward-to: http://169.254.169.254/` static header and asserts it's dropped and never sent.

**Require proxy secret (CodeRabbit Major, cubic P2)** — `use_proxy` now fails unless **both** proxy `url` and `secret` are configured; `X-Proxy-Auth` is always sent (never silently omitted). Centralized in a shared `resolveHttpPushProxy` helper. Added a test for the url-without-secret case.

**Honor proxy for all http_push POSTs (Codex P2)** — `postEphemeralEventToHttpPushNode` now honors `use_proxy` via the same helper, so reactions / read receipts / presence / context updates for a proxied node also route through the proxy. Threaded `httpPushProxy` through `NodeDeliverDeps`/`NodeContextDeps` and every ephemeral deps-construction site. Also fixed that path's own `redirect: 'error'` → `'manual'` (it had the same Workers bug as the durable path).

**Document the wire field (Codex P1)** — Added `use_proxy` to `openapi.yaml` (`HttpPushNodeDelivery`), the TS SDK type (`useProxy`), and README, per AGENTS.md. (Deduped with the bot's openapi addition.)

20/20 conformance tests pass; engine + SDK `tsc` clean.

PR #228, opened 2026-07-01
fix(engine): deliver relayfile inbound integration messages to node agents

khaliqgant:
## What

Lands the **relayfile inbound bridge** (provider → relay agent) on the engine and fixes the one bug that kept it from working end-to-end: **inbound provider messages never reached node/broker-connected agents.**

This is the server-side mirror of the shipped outbound writeback bridge. A provider event (Slack/GitHub/Linear) written to the relayfile VFS fans out to a `webhook_subscription`, which POSTs a signed delivery to the new route `POST /v1/integrations/relayfile/inbound/:workspaceId/:channelId`. The route verifies the HMAC, dedupes, formats the record, and injects it into the bound relay channel.

## The bug

`triggerIntegrationMessage` inserted the message with a **bare insert** and the route then called `fanoutToChannel(…, 'message.created', …)`. Because `message.created` is a `NODE_DELIVERY_EVENT_TYPE`, `fanoutToChannel` deliberately **skips** `sendNodeContextForChannel` for it (normal messages reach nodes via the mailbox/delivery path). Since the route never computed mailbox deliveries either, **node/broker agents received nothing** — only workspace-stream (dashboard/websocket) subscribers did. Given broker 9.1.x node-only delivery, agents got no inbound integration messages at all.

## The fix

- `triggerIntegrationMessage` now writes the message **and** channel deliveries in one atomic unit (mirroring `postMessage` via `buildChannelDeliveryWrite`) and returns `_deliveries` / `_delivery_rejections`.
- The route dispatches those deliveries via `routeDeliveryOutcomes`, run in the **background** (`runInBackground`) alongside the workspace fanout — so a delivery-routing hiccup never fails the ingress. The ingress must return **2xx fast** because the transport retries then DLQs non-2xx.
- Kept the integration insert (not a wholesale swap to `postMessage`) so the **display-author metadata is preserved** — `postMessage` runs `sanitizeUserMessageMetadata`, which would strip the `__relaycast_*` author markers and make messages render as `__relay_webhook__`.
- Surfaced the previously **swallowed** `delivery_failed` error in logs (it hid the root cause during debugging).

## Tests

- New regression test: asserts a **queued delivery row** is created for a channel-member agent after a signed inbound event (the exact thing the bare insert failed to do).
- Full engine suite green: **213/213**. Typecheck + lint clean.

## Notes / follow-ups (not in this PR)

- Binary / base64 / >64KB snapshots still degrade to a skeleton/sliced message (inline snapshot cap is 64KB). Acceptable for now; noted as a follow-up.
- Residual echo consideration: the injected `message.created` re-fires the same channel's outbound writeback subscription; the inbound side already guards `origin === 'agent_write'`. Worth confirming end-to-end.
- Cross-repo: the relay CLI already provisions this route (`integration subscribe` → inbound-target → `webhook_subscription`); relaycast-cloud must have `RELAYCAST_INTERNAL_SECRET` set and be deployed for the live endpoint. E2E (real Slack message → relay channel, no client) is the next gate and needs a preview deploy.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5666,7 +5666,6 @@
       "os": [
         "android"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5688,7 +5687,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5710,7 +5708,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5732,7 +5729,6 @@
       "os": [
         "freebsd"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5754,7 +5750,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5776,7 +5771,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5798,7 +5792,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5820,7 +5813,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -5842,7 +5834,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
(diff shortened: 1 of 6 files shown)


khaliqgant:
Closing as superseded by #229, which contains this same base node-delivery fix plus the subsequent review fixes (5xx on transient failure, dedupe-key requirement), the full ingress hardening pass (byte-exact HMAC, bounded body reader, zod validation, fail-closed idempotency), and an up-to-date merge with main. #229 is the canonical inbound-bridge engine PR.

Issue #213, opened 2026-06-25
Design scoped observer tokens for workspace stream and read-only REST


willwashburn:
## Summary

We should split Relaycast authentication into distinct principals instead of using the root workspace key for both administration and observation.

The immediate design question is: should `/v1/ws?scope=workspace` and read-only history/search REST APIs be accessible with the root `rk_live_` workspace key, or should they use a separate observer credential?

Proposed answer: introduce scoped observer tokens, likely `ot_live_`, and make the workspace stream plus read-only REST surfaces use those tokens. Keep node credentials separate as `nt_live_`, agent credentials as `at_live_`, and reserve `rk_live_` for workspace administration.

## Motivation

Today the workspace stream is effectively a firehose. It is useful for dashboards, brokers, audit tools, backfill systems, and monitoring, but it is too broad to require handing those systems the root workspace key.

A root workspace key can administer the workspace. If that same key is used in browser dashboards, hosted monitors, brokers, or third-party observability tooling, a read-only integration becomes a full workspace compromise.

The new node-delivery direction also makes the boundary clearer:

- Agents should receive through their bound node route.
- Nodes should receive node-scoped delivery/context through node credentials.
- Workspace stream should be observer-only.
- Admin APIs should remain workspace-key-only.

So we should make observation a separate principal, not just another use of the workspace admin key.

### `rk_live_`: workspace admin key

Root/admin credential for a workspace.

Allowed:

- Create, update, delete workspace configuration.
- Register/create agents.
- Create/rotate/revoke agent tokens.
- Create/register nodes.
- Create/rotate/revoke node tokens.
- Create/rotate/revoke observer tokens.
- Configure observer token scopes and filters.
- Administrative listing of tokens, agents, nodes, and workspace settings.

Not ideal long term:

- Opening `/v1/ws?scope=workspace` directly.
- Being embedded into dashboards or monitoring tools.
- Being used by brokers that only need observation.

Migration option:

- Temporarily continue accepting `rk_live_` on the workspace stream for compatibility, but mark it deprecated and add a clear sunset path.
- For a major upgrade, make `/v1/ws?scope=workspace` require `ot_live_` only.

### `at_live_`: agent token

One agent identity.

Allowed:

- Act as exactly one agent.
- Send messages as that agent.
- Read APIs that are already available to that agent.
- Open the agent websocket/direct route.
- Acknowledge/fail/defer deliveries owned by that agent.

Important invariant:

[...]

PR #240, opened 2026-07-08
fix(observer-dashboard): mint scoped observer token for the live stream

willwashburn:
## Summary

Closes the observer-dashboard migration step of the four-principal observer-token model (#213).

Workspace-key (`rk_live_`) logins broke the realtime stream. The engine rejects `rk_live_` on `GET /v1/ws` and accepts only an observer token (`ot_live_`) with `stream:read`, but the dashboard forwarded the login credential straight to the socket. So a workspace-key login got working read-only REST but a **silently dead live stream** — and a comment in the login route incorrectly claimed the stream worked with either credential.

## What changed

- On a **workspace-key login**, the dashboard now mints — or reuses by rotation — a scoped `observer-dashboard` observer token via the admin API and uses it as the stream credential. It's stored in a separate `relaycast_ws_token` cookie so the realtime socket never carries a workspace admin key.
- **Observer-token logins** already work on the stream directly and are unchanged.
- The minted token gets the full read-scope set plus `stream:read` and `include_dms` so the dashboard's firehose view is preserved (the operator already holds the workspace key, so this is a read-only, revocable projection of it — not an escalation).
- A single durable token is kept per workspace: created on first login, rotated on later logins (token material is never retrievable after creation, so reuse = refresh scopes/expiry + rotate). Trade-off noted in code: a login invalidates any other browser still on the previous secret; that session re-syncs on its next session poll or re-login.
- Sessions created before this change fall back to the login credential and re-mint a working stream token on next login. Logout clears the new cookie.

## Files

- `src/lib/observer-token.ts` — new `mintObserverStreamToken` helper (create + reuse-by-rotation, fails soft to null)
- `src/app/api/auth/login/route.ts` — mint on `rk_live_` login, store/return stream token, corrected the misleading doc comment
- `src/app/api/auth/session/route.ts` — return the stored stream token (fallback to login credential for legacy sessions)
- `src/app/api/auth/logout/route.ts` — clear the new cookie
- `src/lib/observer-token.test.ts` — new unit tests

## Validation

- `vitest run` in `packages/observer-dashboard` — 17/17 pass (6 new, covering create, reuse-via-rotate, and every failure path)
- `tsc --noEmit` clean (after building workspace deps)

## Context

The rest of #213 (engine lifecycle/storage/auth, WS + REST scope and filter enforcement, error semantics, all four SDKs, README, and OpenAPI) was already implemented and merged via #214/#222/#230/#231/#232/#233 and is covered by the engine conformance suite. This PR closes the remaining dashboard gap.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kavk2ErRdCo1M4Mo7hCvbw)_

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/observer-dashboard/src/app/api/auth/login/route.ts
+++ b/packages/observer-dashboard/src/app/api/auth/login/route.ts
@@ -4,11 +4,13 @@ import {
   resolveRelayServerCandidatesFromRequest,
   selectEngineForKey,
 } from '../../../../lib/relay-server';
+import { mintObserverStreamToken } from '../../../../lib/observer-token';
 
 export const runtime = 'edge';
 
 const COOKIE_NAME = 'relaycast_key';
 const AGENT_COOKIE_NAME = 'relaycast_agent_token';
+const WS_TOKEN_COOKIE_NAME = 'relaycast_ws_token';
 const ENGINE_COOKIE_NAME = 'relaycast_engine';
 const COOKIE_MAX_AGE = 60 * 60 * 24 * 30; // 30 days
 
@@ -18,13 +20,12 @@ const COOKIE_MAX_AGE = 60 * 60 * 24 * 30; // 30 days
  * fallback), remembers the resolved engine, and sets httpOnly cookies.
  *
  * Accepts either a full workspace key (`rk_live_...`) or a scoped, read-only
- * observer token (`ot_live_...`). Both are treated identically here — the
- * engine enforces the actual read-only scope restrictions for observer
- * tokens on every request, this route just validates the credential and
- * remembers which engine it belongs to. The WebSocket stream authenticates
- * directly with whichever credential was used to log in (workspace key or
- * observer token) — the engine's realtime endpoint (`GET /v1/ws`) already
- * requires an observer token with `stream:read` scope.
+ * observer token (`ot_live_...`). The credential is used for read-only REST
+ * (the engine enforces observer scope restrictions on every request), but the
+ * realtime endpoint (`GET /v1/ws`) accepts *only* an observer token with
+ * `stream:read`. So when the login credential is a workspace key we mint (or
+ * reuse) a scoped observer token here and use that as the stream credential;
+ * an `ot_live_` login already works on the stream directly.
  */
 export async function POST(request: NextRequest) {
   try {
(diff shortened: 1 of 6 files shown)


willwashburn:
Thanks for the reviews (Devin, Codex, cubic, Gemini) — addressed the findings in `61e4956`, which reworks the stream-token strategy rather than patching each symptom.

**P1 — admin key could reach the realtime socket.** Fixed. On mint failure the stream token is now left unset (the socket simply doesn't connect) instead of falling back to `apiKey`. The session route no longer backfills the ws token from the admin key, and the provider passes the observer token (never the admin key) as the socket credential — `wsToken ?? agentToken` can no longer resolve to `rk_live_`. REST reads still use `apiKey`.

**Revoked-name lockout & cross-session invalidation (P2 ×2).** Root-caused to the single fixed-name, rotate-on-login design. Switched to minting a **fresh, uniquely-named** observer token per login (`observer-dashboard-<uuid>`, 30-day expiry). Unique names never collide with the engine's `(workspace_id, name)` unique index, so revoking a token can't permanently lock out re-minting; and because each session gets its own token there's no rotation invalidating other live sessions.

**PATCH response unchecked before rotate (P2).** Moot — the create-only flow has no PATCH/rotate step, so there's no stale-expiry window.

**Logout doesn't revoke on the engine.** Fixed — logout now best-effort `DELETE /v1/observer-tokens/:id` for the token minted this session, so a leaked ws cookie stops working immediately.

**Network error → 500 (earlier round).** Already fixed in `2ef3686`; the mint helper fails soft to `null`.

Tests updated for the create-only mint (returns `{token, id}`) and the revoke helper — 18/18 pass, typecheck clean.

Trade-off worth naming: per-login minting means a workspace can accumulate a few short-lived `observer-dashboard-*` tokens, but they're revoked on logout and expire in 30 days.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #302, opened 2026-07-31
fix(engine): clear left_at when re-resolving a 1:1 DM

khaliqgant:
## What

A 1:1 DM conversation id is a pure function of `(workspaceId, sorted agent pair)`, so it names the durable A↔B relationship rather than a conversation instance.

`resolveConversation` finds an existing 1:1 by joining `dm_participants` with `left_at IS NULL`. If a participant is ever marked departed the `SELECT` misses, the create branch runs, recomputes the **same** deterministic id, and re-inserts with `onConflictDoNothing()` — which no-ops against the existing conversation, channel, and participant rows, leaving `left_at` set.

The conversation then keeps resolving correctly while its roster says someone left. Nothing throws; the two just disagree from then on.

Identifier stability is *not* what breaks — the id never consults `left_at`. What breaks is the claim that the identifier names a relationship with fixed membership.

## Why it matters outside this repo

That claim is load-bearing for the Ratify design partnership. Agent Relay has declared this lifecycle model to Identities AI as the basis on which Ratify delegation certificates are scoped to a DM — invariant **DM-1** in the Agent Relay Resource Identifier Profile. It is stated as an invariant with a test rather than left as a code comment, so it cannot change silently later.

## Change

The create branch clears `left_at` on conflict instead of no-opping. Reaching that branch for an id that already exists means the lookup missed, and for an existing 1:1 the only way it can miss is a departed participant — so re-resolution heals the roster.

Resolution semantics are otherwise unchanged. No current code path can set `left_at` for a 1:1: it is written only by the group-DM code (`groupDm.ts`). The test reaches that state directly so the invariant is guarded before some future path can reach it accidentally.

## Tests

Four tests covering the model:

- id derives from `(workspace, sorted pair)`
- either sender resolves the same conversation
- re-resolution creates no second conversation
- **DM-1**: a 1:1 conversation never has a departed participant

The DM-1 test fails against the previous behaviour, which is the point — it pins the invariant, not the current behaviour.

## Verification

- `npx turbo build && npx vitest run` in `packages/engine` — **515/515**
- `npx turbo lint` — clean
- Confirmed it is a real guard: with the one-line fix reverted, DM-1 is the only test in the engine suite whose result changes.

## Review note

This touches production message-routing code (`onConflictDoNothing` → `onConflictDoUpdate` in the 1:1 resolution path). It is one line and only fires in a branch unreachable today, but a reviewer who owns this file should confirm the conflict-target choice.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,11 @@ This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 Packages without a separate changelog are covered by the cross-package notes below.
 
-## [Unreleased]
+## [Unreleased - Patch]
+
+### Fixed
+
+- Sending into a 1:1 DM whose participant is marked as departed now restores that participant, instead of resolving the conversation while its roster still shows the departure.
 
 ## [6.3.0] - 2026-07-28
 
--- a/packages/engine/CHANGELOG.md
+++ b/packages/engine/CHANGELOG.md
@@ -7,7 +7,11 @@ See the [root changelog](../../CHANGELOG.md) for cross-package release highlight
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [Unreleased - Patch]
+
+### Fixed
+
+- `sendDm` re-resolving a 1:1 conversation now clears `dm_participants.left_at` for both participants instead of no-opping on conflict. A 1:1 with a departed participant previously resolved to the same conversation while its roster still showed the departure; the conversation id is unchanged either way, since it derives only from the workspace and the sorted agent pair.
 
 ## [6.3.0] - 2026-07-28
 
--- a/packages/engine/src/engine/__tests__/dm.test.ts
+++ b/packages/engine/src/engine/__tests__/dm.test.ts
@@ -1,0 +1,172 @@
+/**
+ * 1:1 DM conversation identity and roster invariants.
+ *
+ * A 1:1 conversation id is a pure function of `(workspaceId, sorted agent pair)`,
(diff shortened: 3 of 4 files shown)


khaliqgant:
Addressed the Codex P1 in 2a8f23a — good catch, the AGENTS.md release-level rule was a repo convention I missed rather than an oversight I'd argue with.

- Root `CHANGELOG.md`: bullet under the existing `[Unreleased - Patch]` → `### Fixed`.
- `packages/engine/CHANGELOG.md`: raised the bare `[Unreleased]` heading to `[Unreleased - Patch]` and added the package-level detail.

Patch is the right level: behaviour only changes in a state no current code path can reach (`left_at` is written solely by the group-DM path), and the conversation id is unchanged either way since it derives only from the workspace and the sorted agent pair.

CodeRabbit reported no actionable comments.

khaliqgant:
Rebased onto `main` (was based on a stale checkout two commits behind, which is what made this conflict once the changelog edits landed). Now at `7a88f50`, mergeable, CI green.

The changelog entries needed re-filing after the rebase: `v6.3.0` had already been cut, so the bullet I'd added would have landed inside the *released* section. Both now sit under the pending heading:

- root `CHANGELOG.md`: raised `[Unreleased]` → `[Unreleased - Patch]` with the user-facing bullet
- `packages/engine/CHANGELOG.md`: same heading change, plus the package-level detail

Re-verified on the new base: `turbo build` + engine suite **524/524**, `turbo lint` clean.

PR #318, opened 2026-08-09
feat(a2a): federate Ratify proofs and revocations

khaliqgant:
## Summary

- make standard bare agent-card discovery work for a sole workspace while keeping multi-tenant discovery fail-closed
- make an explicit path workspace beat host inference so the documented path route works on multi-label authorities
- define version 1 Ratify proof-bundle and signed-revocation metadata under com.agentrelay.ratify
- preserve structured DM metadata across A2A send, receive, webhook, durable delivery, public types, SDK, and OpenAPI
- authenticate inbound registered peers and prevent their tokens from becoming an open A2A relay
- add a credential update step that completes reciprocal A2A registration without rotating either exchanged token

## Discovery decision

Option (b), sole-workspace fallback, is the actual standards fix because it makes the bare /.well-known/agent-card.json path work without Relaycast-specific query or path conventions. The fallback only runs without an explicit selector and only when exactly one workspace exists. Hosted subdomain selection remains authoritative and is covered by a test.

The explicit path selector was also moved before host inference. That is route correctness, not a substitute for option (b): without it, the documented /:workspace/.well-known/agent-card.json route remains dead on every conformant multi-label authority. Invalid explicit selectors do not fall through to another tenant.

This production discovery defect is independently deployable and deserves its own expedited PR. It is included here because it is on the federation critical path; the discovery hunks/tests can be split or cherry-picked if maintainers want separate release cadence.

## Wire, authentication, and trust

The interface and reciprocal three-step credential handshake are documented in docs/a2a-ratify-federation.md. The conformance test sends the full proof A-to-B and the signed revocation stream B-to-A, with a distinct registration-issued bearer token authenticating each direction.

The named test blocks unauthenticated and unregistered peers from injecting Ratify proof metadata. It checks both a request with no credential and a workspace key that is valid for management but is not a registered peer token; neither creates a local DM. Registered peer tokens also cannot relay to another external A2A target.

Revocations are applied by the verifier only after issuer_id resolves to a pre-trusted hybrid public key, the carried key matches it, and Ratify verifyRevocationList succeeds. A carried public key is never treated as its own trust anchor.

## Measurements

- full proof bundle: 131,072 UTF-8 bytes crossed two isolated Relaycast engine instances end to end
- resulting forwarded JSON-RPC body: 131,534 bytes
- 131,073-byte bundle: rejected before egress
- revocation issue-to-observable-refusal, n=20: median 14.02 ms, p95 16.66 ms, max 25.78 ms
- with 75 ms injected transport delay: observable refusal at 96.27 ms

The engine path is measured. The two in-memory deployments preserve the complete HTTP, authentication, JSON-RPC, persistence, signature-verification, and refusal path, but do not include real DNS, TLS, or a production reverse proxy. The production ingress body cap is not measurable without deploying or production A2A credentials, both outside this task constraints; the real two-host rehearsal must confirm that edge separately.

## Negative controls

I temporarily removed each protection, rebuilt first, and confirmed the relevant tests turned red:

- path-before-host selector ordering
- exactly-one-workspace fallback guard
- 128 KiB proof limit
- registered-peer inbound authentication gate
- revocation signature verification
- reciprocal auth-credential persistence

The original mutation run produced seven expected failures. Removing reciprocal credential persistence additionally made the reverse-direction revocation test fail 401. Every guard was restored.

## Verification

- npx turbo build: 9/9 tasks passed, engine remains 7.0.0
- npx turbo lint: 13/13 tasks passed
- focused Vitest: 153/153 passed
- TruffleHog committed-diff scan: 0 verified and 0 unverified secrets
- GitHub CI on the amended commit passed Lint, Build & Test and Rust SDK
- direct root Vitest invocation: 1,414 passed; 11 pre-existing failures plus one pre-existing suite-load failure remain in the known React DOM-environment and types test setup areas

## Live-run blockers

- No npm publish or deployment was performed. A stock published 7.0.0 engine image cannot contain this source fix; build the self-host image from this commit/PR or wait for a future published artifact.
- Production edge request limits and real DNS/TLS latency still require a two-deployment rehearsal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 Packages without a separate changelog are covered by the cross-package notes below.
 
-## [Unreleased - Minor]
+## [Unreleased - Major]
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Packages without a separate changelog are covered by the cross-package notes bel
 ### Added
 
 - Added a pinned, multi-architecture self-host container and Compose runbook with persistent SQLite/files state, health checks, and mandatory HTTPS deployment-authority validation.
+- A2A federation now preserves structured DM metadata end to end, defines a versioned `com.agentrelay.ratify` carrier for proof bundles and signed revocation lists, and supports reciprocal authenticated delivery between registered peers.
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -647,6 +647,7 @@ A2A (Agent-to-Agent) gateway endpoints:
 ```text
 POST   /v1/a2a/register              Register an external A2A agent
 GET    /v1/a2a/agents                List registered A2A agents
+PATCH  /v1/a2a/agents/:name          Complete or rotate an A2A connection
 DELETE /v1/a2a/agents/:name          Remove an A2A agent
 GET    /v1/a2a/agents/:name/card     Get agent card for a registered agent
 GET    /.well-known/agent-card.json  A2A agent card (root-level; ?workspace= selects on multi-tenant)
@@ -654,6 +655,10 @@ GET    /:workspace/.well-known/agent-card.json  A2A agent card for a named works
 POST   /a2a/rpc                      A2A JSON-RPC gateway (root-level)
 POST   /a2a/webhook/:ws/:name        Inbound webhook for relay agents
 ```
+
+Single-workspace deployments serve the standard bare agent-card URL without a
+workspace query parameter. A2A messages can carry versioned Ratify proof and
+revocation metadata; see [Ratify over A2A](docs/a2a-ratify-federation.md).
 
 Programmability, directory & observability:
 
--- a/docs/a2a-ratify-federation.md
+++ b/docs/a2a-ratify-federation.md
@@ -1,0 +1,99 @@
+# Ratify over A2A
+
+Relaycast carries Ratify Protocol proofs and revocations in the A2A message
+metadata key `com.agentrelay.ratify`. The key lives at
+`params.message.metadata["com.agentrelay.ratify"]`, not in the gateway's
+top-level `params.metadata` routing fields.
+
+The version 1 payload is a closed, discriminated shape. Wire field names are
+snake_case. A receiver must reject an unknown `version` or `kind` rather than
+guessing how to interpret it.
+
+## Proof bundle
+
+```json
+{
+  "version": 1,
+  "kind": "proof_bundle",
+  "correlation_id": "challenge-or-invocation-id",
+  "bundle": "{canonical Ratify ProofBundle JSON}",
+  "grant": "{optional canonical Ratify DelegationCert JSON}",
+  "operation": {
+    "invocation_id": "task-42"
+  },
+  "task": {
+    "title": "Update the runbook",
+    "instructions": "Edit the deployment section",
+    "path": "docs/"
+  }
+}
+```
+
+`bundle` is required in both presentation directions. `grant`, `operation`,
+and `task` are optional so the same carrier supports a mutual presentation or
+a delegated task handoff. When present, `grant` is the canonical Ratify
+DelegationCert wire JSON. The UTF-8 byte length of `bundle` may not exceed
+131,072 bytes (`MAX_PROOF_BUNDLE_BYTES`, 128 KiB). The surrounding A2A JSON-RPC
+request is necessarily larger and must not be capped at 128 KiB by a proxy.
+
+## Revocation list
+
+```json
+{
+  "version": 1,
+  "kind": "revocation_list",
+  "issuer_id": "human:northwind.example:alice",
+  "updated_at": 1786310000,
+  "revoked_certs": ["cert_01"],
+  "issuer_pub_key": {
+    "ed25519": "base64...",
+    "ml_dsa_65": "base64..."
+  },
+  "signature": {
+    "ed25519": "base64...",
+    "ml_dsa_65": "base64..."
+  }
+}
+```
+
+The receiver must not apply any `revoked_certs` merely because this metadata
+arrived over an authenticated A2A connection. It must:
+
+1. resolve `issuer_id` to an already trusted issuer public key;
+2. require `issuer_pub_key` to match that trusted key;
+3. reconstruct the Ratify `RevocationList` from `issuer_id`, `updated_at`,
+   `revoked_certs`, and `signature`; and
+4. call Ratify `verifyRevocationList` with the trusted issuer key before
+   changing local revocation state.
+
+The public key beside a signature is not its own trust anchor. A self-signed
+attacker payload that is not bound to the expected `issuer_id` must fail closed.
+
+## Two-Relaycast handshake and delivery
+
+For deployments A and B to exchange messages in both directions:
+
+1. B registers A's agent card with `POST /v1/a2a/register`, setting the skill
+   on A as `target_agent`. B returns a relay proxy name and bearer token for A.
+2. A registers B's card, stores B's returned token as `auth_credential`, and
+   sets the skill on B as `target_agent`. A returns its proxy token for B.
+3. B completes the reciprocal connection with
+   `PATCH /v1/a2a/agents/{a-proxy-name}`, setting A's returned token as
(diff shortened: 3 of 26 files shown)


khaliqgant:
Addressed the P1s and the 🔴 in `0ae314b`, and split the discovery work out so this PR is now federation-only.

**Devin 🔴 / cubic P1 — SDK rewrote inner field names of `data`/`metadata`.** Correct, and the consequence is worse than untidy: a `RevocationList` whose `revoked_certs` arrives as `revokedCerts` cannot be reconstructed byte-for-byte, so its signature fails and a cross-deployment revocation is rejected **for the wrong reason** — the kill switch appears to work while proving nothing. Both keys now join `VERBATIM_VALUE_KEYS` next to `headers`, `input` and `input_schema`, which were already exempt on precisely this reasoning.

⚠️ **This changes behaviour for existing consumers** and I have not hidden it in the changelog: callers who relied on the SDK snake_casing their metadata keys on the wire will now see keys exactly as written, **including agent-registration metadata**. The `setup.test.ts` assertion expecting `favorite_color` encoded the corruption rather than the contract — a caller who deliberately wrote `favoriteColor` got a different key back — so I corrected the expectation rather than preserving it. That is the one judgement call here worth a second opinion.

**Codex P1 — retried inbound messages delivered twice.** Correct. The inbound handler called `sendDm` directly and skipped the idempotency the DM route already applies to every send, while the counter, webhook, workspace event and delivery routing all run *after* the durable write. Inbound now goes through `runIdempotent`, keyed on the caller's `message_id` (falling back to the JSON-RPC id) and scoped to the registered caller so two counterparties cannot collide. Replays skip the side effects and return the original result, so a retry is observably identical to a single call.

**cubic P1 — Swift dropped the metadata.** Correct; `CoreMessagePayload` decoded successfully and silently lost the field. Added as optional `[String: JSONValue]`, mirroring `CoreMessagePayloadSchema`.

## Both new tests are non-vacuous

Checked rather than assumed:
- reverting **only** the casing change fails the two round-trip assertions
- disabling **only** the idempotency key delivers the retried message twice

## Now disjoint from #319

The agent-card discovery fix has been removed from this branch and ships on its own in #319, which is where its review lives. `packages/engine/src/routes/a2a.ts` here carries federation code only.

## Suites

```
engine            53 files / 556 tests
sdk-typescript    419 tests
turbo build       green
```

Remaining open here are the P2/P3 items (mailbox limits on inbound A2A DMs, the observer-history and idempotency-fingerprint notes, schema nits). Working through those next.

khaliqgant:
Follow-up on the casing change — I flagged it as breaking, then measured it instead of leaving that as an assertion. `616b300`.

**What can actually change:** only **multi-word** keys, and only for callers reading or writing `data`/`metadata` through this SDK. Single-word keys (`source`, `fleet`, `release`, `cli`) are identical in both casings.

**What is in production:** 1025 agents, 723 with metadata. Eight multi-word keys in use — `node_id` (117), `registered_at` (117), `cloud_workspace_id` (20), `cloud_agent_id` (19), `invocation_id` (15), `identity_key` (9), `room_membership_id`, `room_identity_id` — all stored snake_case. So an SDK reader sees `nodeId` today and `node_id` after.

**What reads them:** grepped `relay`, `relaycast`, `cloud` and the VS Code extension for the camelCase forms (`metadata.nodeId`, `metadata.cloudWorkspaceId`, …) — **zero hits**. The only consumer that reads agent metadata at all is `cloud/packages/web/lib/room/relaycast-access.ts`, which checks `metadata.cloud_workspace_id` in snake_case and reaches the API through `globalThis.fetch` rather than this SDK. It reads the wire format directly and is unaffected either way.

So the blast radius inside our own systems is zero. The residual is an external SDK consumer reading multi-word metadata keys — and for anyone who deliberately wrote snake_case keys this is a fix, since they were already getting a different key back than they wrote.

The changelog entry now says that, rather than leading with a bare "Breaking".

khaliqgant:
**cubic P2 — pending release level.** Accepted, raised to `[Unreleased - Major]` on the root and sdk-typescript changelogs in `ad6b01d`. Engine and types stay Minor; neither changes behaviour for existing callers.

Worth recording that there is precedent pointing the other way, since silently overriding it would be the wrong way to resolve this. **6.2.0 shipped the identical class of change as a Minor, framed as a Fix** — that release stopped key-rewriting inside `input_schema`/`output_schema`, `input`/`output` and `headers`, on the reasoning that rewriting user-authored keys was corruption rather than contract. By that precedent this entry is Minor too: `data`/`metadata` are the same kind of value and were simply missed from the same set.

Taking Major anyway, because the two differ in one material way. `input`/`output` and `headers` are action payloads whose reader is essentially the code that produced them, so a rename is self-consistent end to end. Agent and message metadata is written by one system and read by another — fleet and cloud write `node_id`, `registered_at`, `cloud_workspace_id`; separate code reads them — and a silent rename across a system boundary is exactly where this bites. A behavioural change on a cross-system field-name contract should not ride a minor bump.

Happy to be overruled back to Minor on the precedent; flagging the reasoning so it is a decision rather than an accident.

khaliqgant:
Remaining P2/P3 items addressed in `d45ff18`. `turbo test` 18/18.

**Mailbox config** (raised twice, same defect) — inbound A2A passed only `skipA2aIntercept`, so `sendDm` fell back to its fixed one-hour / 1000-message defaults. Registered-peer delivery was the one path on the deployment exempt from its own configured TTL and backpressure. Now resolves and passes the config as `/v1/dm` does.

**Workspace/observer DM history dropped metadata** — `getDmMessagesForWorkspace` never selected or projected `messages.metadata`, so a federated delivery's Ratify envelope was visible to the recipient and invisible in history: the same message reading differently depending on which door you came through. Projected through `publicMessageMetadata`.

**Idempotency fingerprint size** — `data` was embedded whole. At 128 KiB per proof bundle that put ~256 KiB into the KV record for the TTL and compared the full payload on every replay. Now digested with the `sha256Hex` the middleware already uses; a digest answers the only question the fingerprint asks, in constant size.

**Inbound acknowledgement latency** — fanout, delivery routing and rejection notification were awaited, so the counterparty's "message accepted" waited on our recipient's delivery, including a slow HTTP-push receiver. Moved to `runInBackground`, matching `/v1/dm`. The durable write is what the response attests to.

**`message/stream`** — accepted, then served as a one-shot returning an already-terminal task with no stream channel. A caller expecting incremental updates would silently believe a truncated conversation was still open. Now refused with `-32601` naming `message/send`, so it falls back immediately. ⚠️ **The agent card still advertises `message/stream`** — reconciling that is wider than this PR, so it is flagged rather than changed here.

**Base64 validation** — `issuer_pub_key` and `signature` are documented base64 but only checked non-empty, so a value that could never decode passed at the edge and failed later inside the verifier as an opaque error. Now validated. The existing fixtures used `'pub-ed'`/`'sig-ed'`, which are not base64 — replaced with real values, and added cases covering the rule including a valid one that must still pass.

**Proof-bundle size guard (P3)** — the whole string was encoded before deciding it was oversized, so an attacker could force a multi-megabyte encode plus a second same-size allocation before rejection, on unauthenticated inbound traffic. String length is checked first; UTF-8 cannot shrink, so length over the cap already proves byte length over it.

**Revocation test** — you are right that it validated a test-local helper, though the engine is not the enforcement point (a Ratify verifier is), so there is no production enforcement here to drive. The property the engine *does* own is that a signed document crosses unmodified, so the test now asserts the received wire deep-equals the sent one in addition to verifying its signature. A key rename fails with a readable diff instead of an opaque signature failure — which is exactly how the casing bug in this PR would have surfaced.

khaliqgant:
**cubic P1 — unpadded base64 rejected.** Correct, and it is a regression I introduced while hardening this very field. Requiring a multiple-of-four length rejects unpadded base64, which is legal and commonly emitted — and the value it would reject is a **signed revocation list**. Over-strict validation there fails the kill switch closed over a formatting preference, which is a worse failure than the loose check it replaced.

Fixed in `92a6a6f`: padding is optional. A padded value must still be a multiple of four; an unpadded one is rejected only at length ≡ 1 (mod 4), which no base64 quantum can produce. Tests now cover `QUJDRA` and `QUJDRA==` — the same bytes, both accepted — alongside the malformed cases.

Worth noting for the record: this is the second time in this PR that a change intended to protect the revocation path would have broken it. The first was the SDK casing transform corrupting `revoked_certs`; this one was my own validation. Both fail the same way — the kill switch refuses for a reason unrelated to authority — which is exactly the failure the demonstration cannot afford, because it looks identical to the protection working.

PR #319, opened 2026-08-10
fix(a2a): serve agent-card discovery at the standard well-known path

khaliqgant:
Split out of the Phase 2 federation work (#318) so it can ship on its own — it affects any A2A counterparty, not just that engagement.

## Current production behaviour

Agent-card discovery is where A2A federation starts, and it does not work today on any deployment whose workspace is not named after the first label of its hostname. Verified against `cast.agentrelay.com`, whose workspace is actually named `rw_7ccfea89`:

```
GET /.well-known/agent-card.json                        -> 404 workspace_not_found
GET /.well-known/agent-card.json?workspace=rw_7ccfea89  -> 200
GET /rw_7ccfea89/.well-known/agent-card.json            -> 404 workspace_not_found
GET /.well-known/agent-card.json  (Authorization: Bearer …)  -> 200
```

## Two defects, one root cause

`extractWorkspaceHint` resolved **query → host label → path param**, and the unauthenticated card handler had no other source.

1. **The documented `/:workspace/.well-known/agent-card.json` route was dead** on any authority with three or more labels — host inference always answered first, so the path parameter was never read. The Relay identifier profile *requires* multi-label, non-loopback authorities, so every conformant deployment sits exactly where that escape hatch cannot work.
2. **A single-tenant self-host could not answer the bare standard URL at all** — the one an A2A counterparty tries first.

`POST /a2a/rpc` is unaffected; it authenticates and takes the workspace from the token. That is what made this narrow and misleading: discovery 404s while everything authenticated succeeds, and the error text blames agent registration.

## The fix

- An explicit path selector now beats host inference.
- The card handler falls back to the sole workspace when a deployment has exactly one.
- The fallback is **skipped whenever a query or path selector was supplied**, so a misspelled selector fails loudly instead of quietly resolving to another tenant.
- The candidate query is bounded to , so a multi-tenant deployment fails closed without scanning the workspace table.

## Regression risk, and the test that covers it

This reorders **live routing**. The test that makes it safe to ship is *"keeps a valid hosted workspace subdomain authoritative"* — hosted multitenancy behaviour is unchanged.

Six tests total; the two subtle ones assert that the fallback never papers over a caller error.

## Verification

- `packages/engine` full suite: **558 passed (558)**, 52 files — no regressions
- `a2a.test.ts`: 33/33

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,11 @@ This project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 Packages without a separate changelog are covered by the cross-package notes below.
 
-## [Unreleased]
+## [Unreleased - Patch]
+
+### Fixed
+
+- A2A agent-card discovery now answers the standard `GET /.well-known/agent-card.json` on a single-tenant deployment, and an explicit `/:workspace/` path selector is honoured instead of being overridden by host inference. Previously a deployment whose workspace was not named after the first label of its hostname returned `workspace_not_found` at the standard discovery URL, and the documented path route could never take effect on any authority with three or more labels. Multi-tenant deployments still fail closed, and an invalid explicit selector is never silently replaced.
 
 ## [7.0.0] - 2026-08-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Packages without a separate changelog are covered by the cross-package notes bel
 
 ### Fixed
 
-- A2A agent-card discovery now answers the standard `GET /.well-known/agent-card.json` on a single-tenant deployment, and an explicit `/:workspace/` path selector is honoured instead of being overridden by host inference. Previously a deployment whose workspace was not named after the first label of its hostname returned `workspace_not_found` at the standard discovery URL, and the documented path route could never take effect on any authority with three or more labels. Multi-tenant deployments still fail closed, and an invalid explicit selector is never silently replaced.
+- A2A counterparties can discover a self-hosted deployment at the standard `GET /.well-known/agent-card.json`, and `/:workspace/.well-known/agent-card.json` now resolves. Deployments holding more than one workspace still require a selector.
 
 ## [7.0.0] - 2026-08-07
 
--- a/README.md
+++ b/README.md
@@ -649,7 +649,8 @@ POST   /v1/a2a/register              Register an external A2A agent
 GET    /v1/a2a/agents                List registered A2A agents
 DELETE /v1/a2a/agents/:name          Remove an A2A agent
 GET    /v1/a2a/agents/:name/card     Get agent card for a registered agent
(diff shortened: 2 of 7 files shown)


khaliqgant:
Review addressed in `ada0cff`. Engine suite **559/559** across 52 files.

**Devin — docs parity.** Correct, and it is a repo rule I missed. `openapi.yaml` now documents the resolution order, the `?workspace=` selector, the path-scoped route as its own path item, and the `404 workspace_not_found` response on both forms. `README.md` lists the path-scoped route.

**Devin — sole-workspace fallback firing after an unresolved host label.** Keeping this, deliberately, and the reasoning is now in the code rather than implied.

The guard distinguishes two things that look alike. An explicit `?workspace=` or `/:workspace/` is *caller intent*, so a typo must 404 rather than resolve to another tenant. Host-label inference is not caller intent — it is the hosted workspace-per-subdomain convention, and on any authority with three or more labels it **always** yields a candidate. The Relay identifier profile requires multi-label authorities, so treating an inferred label as an explicit selector would mean the fallback never fires on precisely the deployments it exists for. That would reinstate the bug this PR fixes.

The row cap is the real boundary: with two or more workspaces the fallback declines rather than guessing, so there is no tenant boundary to cross. On a deployment holding exactly one workspace there is only one tenant, and the card is unauthenticated by design — that is what A2A discovery is.

Rather than leave that as an argument, it is now asserted both ways:
- `serves the sole workspace even when an unresolved host label was inferred` — new, states the single-tenant behaviour as intent
- `fails closed on a bare multi-tenant deployment instead of selecting a workspace` — pre-existing, and it is the *same request shape* (host label `cast` against workspaces named `northwind`/`borealis`), returning 404

**CodeRabbit — changelog concision.** Fair; both entries were carrying implementation backstory. Trimmed to one impact-first bullet each.

khaliqgant:
**CodeRabbit — `workspace` is not required on every multi-tenant deployment.** Correct; the description was wrong. A valid workspace subdomain resolves on its own, as the integration test shows (`northwind.cast.example/.well-known/agent-card.json` → 200, no query parameter).

Fixed in `4349727`. The endpoint description had the same ambiguity — "unless a selector identifies it" read as excluding the host label, which is one of the four resolution mechanisms listed a line above it. Both now say a valid workspace subdomain is sufficient on its own, while keeping the part that matters: an explicit query or path selector that does not resolve still 404s rather than falling back to another workspace.

PR #324, opened 2026-08-13
fix(auth): enforce sponsor authority server-side

khaliqgant:
## Security impact

Closes the workspace-key sponsor-impersonation authority gap identified while reviewing AgentWorkforce/relay#1497. Hosted Relaycast now verifies the signed RelayAuth sponsor grant at the token-issuance authority instead of trusting a client wrapper or editable agent metadata.

## What changed

- Verify RS256 sponsor grants with a pinned public key, RFC 7638 `kid`, exact issuer/audience/org, `identity.create` intent, token type, lifetime, sponsor id, and OIDC subject.
- Require authority before hosted workspace creation, REST agent registration/rotation, fleet `agent.register`, and A2A proxy credential issuance.
- Persist sponsor, OIDC identity, and work-unit ownership in server-controlled, write-once columns plus a durable `(workspace, agent name)` claim that survives agent deletion.
- Require matching authority for destructive deletion/release, including generic and node-addressed built-in release action aliases, so callers cannot bypass the dedicated route and delete-then-recreate cannot transfer a protected name.
- Add an incumbent-agent-token-only one-time migration route for legacy rows. A workspace key cannot invoke it.
- Reject sponsor grants whose signed lifetime exceeds RelayAuth’s 15-minute issuance maximum.
- Recheck the complete durable name claim at the database mutation boundary so a stale concurrent preflight decision cannot win a delete/recreate race.
- Reject credential-authority keys in caller-editable metadata.
- Add Rust SDK authority request methods and wire types.

## Security tests

`agentCredentialAuthority.test.ts` proves:

- a raw `POST /v1/agents` with only a workspace key is rejected;
- rewriting sponsor/identity metadata cannot rotate or reclaim another credential;
- invalid/expired proofs create no workspace or key side effect;
- legacy binding requires the incumbent agent token, cannot be rebound, and a rejected D1 node admission cannot poison its pending durable claim;
- node-control `agent.register` cannot bypass authority;
- delete/recreate cannot transfer a protected name.

## Validation

- `npx vitest run packages/engine/src --maxWorkers=1` — 54 files / 572 tests passed
- `npm test -w @relaycast/types` — 163 tests passed
- `npm run build -w @relaycast/engine`
- `npm run lint -w @relaycast/engine`
- Rust authority wire test passed
- Cloudflare workerd integration against the exact packed engine: 37/37 passed against exact packed commit `a87e6e7f`
- `cargo check --manifest-path packages/sdk-rust/Cargo.toml` passed

A highly parallel root `npm test` run also exposed worker-start/test-isolation timeouts in unrelated SDK/MCP tests. The engine and types suites above were rerun serialized and passed with all assertions green.

## Cloudflare portability finding

A real workerd run exposed that jose imports public verification keys as non-extractable by default, so exporting the JWK for RFC 7638 `kid` validation returned `credential_authority_unavailable` on Cloudflare. Commit `0b29cc36` imports the public verification key as extractable (public material only); the full workerd suite now verifies legitimately signed grants.

## Deployment dependency

This is intentionally draft. Hosted enforcement is not active until relaycast-cloud is updated to bind the RelayAuth public key/issuer and consumes a release containing this engine change. Cloud integration is staged in AgentWorkforce/relaycast-cloud#60, which remains blocked on publishing these packages and updating its lockfile. No deployment is performed by either PR.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/324?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


miyaontherelay:
## Open question for the RelayAuth contract owner: sponsor grants are replayable within their TTL

Reviewed as relaycast-324 lane. Filing this as an explicit open question rather than a passing note, per lead direction — this needs a decision, not an assumption.

**Mechanism** (`packages/engine/src/engine/agentCredentialAuthority.ts`):
- `jti` (`grantId`, line 138) is extracted from the JWT and required to be present/non-empty (line 157: `!grantId` rejects), but is **never persisted or checked against a store anywhere in the engine** (verified: no `jti`/replay tracking exists outside this file). A syntactically valid, signed sponsor grant can be replayed any number of times until its own `exp` (max 15 min, `MAX_GRANT_TTL_SECONDS`, line 30/160).
- The claim that would bind a grant to *one specific identity* — `work_unit_key` — is **not part of the signed JWT at all**. It's caller-supplied, unsigned, free text (`packages/types/src/registration-authority.ts:15`, `z.string().min(32).max(4096)`), passed alongside the proof and hashed into the binding client-side of verification (`toAgentBinding`, line 209-223).

**Consequence:** `intent: 'identity.create'` (`REQUIRED_INTENT`, line 25) reads like a per-action authorization claim, but it is not enforced as one. A single sponsor grant authorizes an *unbounded* number of `identity.create` actions (different agent names, different `work_unit_key` values) for its full 15-minute lifetime — it authorizes "sponsor S may create identities" as a standing capability for the TTL window, not "sponsor S may create *this* identity, once."

This does **not** cross the sponsor boundary this PR closes (every resulting agent is still correctly bound to the same verified sponsor/org/OIDC subject — confirmed via the cross-sponsor test at `agentCredentialAuthority.test.ts:65-192`). But it is a real gap between what the claim name implies and what is enforced, and it should be a deliberate decision, not an artifact of `jti` being validated for presence and then discarded.

**Two options, either of which closes it:**
1. Persist `jti` (e.g. a short-lived dedup table/cache keyed on `jti`, TTL ≤ 15 min) and make grants single-use — reject on second presentation.
2. Move `work_unit_key` (or a hash of it) **into the signed claims** issued by RelayAuth, so a grant is cryptographically bound to one identity/work-unit rather than trusted from an unsigned caller-supplied field.

If unbounded batch issuance from one grant is intentional (e.g. a sponsor authorizing a fleet-wide provisioning run), that's a legitimate design — but it should be stated as such here, not left implicit.

— relaycast-324

miyaontherelay:
## `RESERVED_METADATA_KEYS` should be enforced at the metadata write boundary, not at two named call sites

Reviewed as relaycast-324 lane. Enumerated every write to `agents.metadata` in the engine:

**Guarded** (caller-supplied keys are spread directly into the column, and `assertNoCredentialAuthorityMetadata` is called first):
- `routes/agent.ts:262` — `POST /v1/agents` (register)
- `routes/agent.ts:361` — `PATCH /v1/agents/:name` (update)

**Unguarded, but currently safe only because the caller-supplied data never becomes a top-level metadata *key*** (values only, or fully server-constructed key shapes):
- `engine/node.ts:1206` / `:1221` — fleet `agent.register`, hardcoded `{ fleet: { node_id, invocation_id, registered_at } }`
- `engine/a2a.ts:321-327` — A2A proxy registration, fixed-key `proxyMetadata` object built from parsed agent-card fields
- `engine/action.ts:769` — release/delete, `json_patch` of a fixed `{ release: { reason, released_at, previous_name } }` shape, where `reason` is the only caller input and it lands as a *value*, never a key

**The problem:** this safety is structural/incidental today, not enforced. Nothing prevents a future endpoint from spreading caller-supplied metadata *keys* (the same pattern the two guarded sites already use) without remembering to call `assertNoCredentialAuthorityMetadata` first. If that happens, the reserved-key protection silently doesn't apply to the new sink, and no existing test would catch it — the two current tests only exercise the two current call sites.

**Recommendation:** move the check to the single place every `agents.metadata` write passes through (e.g. a thin wrapper around the update/insert helpers in `engine/agent.ts` that all these call sites already funnel through, or a Drizzle-level guard on the `metadata` column write) rather than sanitizing at named call sites. Sanitizing two sinks relocates the problem to the third one that gets written next, rather than closing it.

— relaycast-324

miyaontherelay:
## Destructive-release authority is enforced by convention at call sites, not inside the function that performs the mutation

Reviewed as relaycast-324 lane. Ran the inverse survey (every mint/rotate/delete of an `agents` row, independent of the four authority functions) rather than only grepping authority-function call sites, since that search can't find a route that *should* call authority and doesn't.

**Mint/rotate reconcile cleanly.** Every `INSERT`/`tokenHash` write into `agents` traces back to a gated caller, with one benign exception: `engine/inboundWebhook.ts:28` (`ensureWebhookAgent`) mints a token with no authority check, but for the hardcoded system-sentinel name `__relay_webhook__` — never caller-supplied, idempotent via `onConflictDoNothing`, can't rotate or steal an existing row.

**Destructive release does not reconcile.** `dispatchRelease` (`engine/action.ts:664`) and `applyReleaseCompletionEffect` (`engine/action.ts:1302`) — the two functions that actually tombstone-rename+rotate-token (`:702`, `:758-766`) or hard-`delete(agents)` (`:1361`) — contain **no credential-authority check themselves**. They trust `input.name`/`input.delete_agent` read directly from the invocation's stored input. Authority is enforced entirely by the *caller* of `invokeAction`/`invokeNodeAction`.

I found 5 callers of those two entry points:
- `routes/agent.ts:686` (`POST /v1/agents/release`) — gated at `:672`
- `routes/action.ts:193` (generic invoke) — gated at `:183` (`actionName==='release' && input.delete_agent===true`)
- `routes/node.ts:295` (node-addressed invoke) — gated at `:287`, same condition
- `routes/agent.ts:472` (`POST /v1/agents/spawn`) — `actionName` is the hardcoded literal `'spawn'`, can't become `'release'`
- **`engine/trigger.ts:192` (`fireMessageTriggers`) — not gated at all.**

`fireMessageTriggers` calls `invokeAction(db, workspaceId, row.actionName, { input: { trigger_id, message } }, ...)`. `row.actionName` is set via `POST /v1/triggers` (`routes/trigger.ts:22`: `action_name: z.string().min(1)`, no allowlist, `requireAuth` only). **Nothing stops a caller from creating a trigger with `action_name: 'release'`**, which then fires unauthenticated on every matching message straight into `dispatchRelease`.

**Not exploitable today**, and it's worth being precise about why: `fireMessageTriggers`'s input is hardcoded to exactly `{trigger_id, message}` — no top-level `name` key ever. `dispatchRelease` 400s immediately without `input.name` (`:676-679`), and the destructive `completeLocally()` branch additionally requires `input.delete_agent === true` (`:879`, `:897`) — a field the trigger path never sets. So a trigger wired to `action_name: 'release'` today fires and immediately fails closed, silently swallowed by `trigger.ts`'s own catch-and-log (`:204-209`).

**Why this is worth fixing anyway:** the safety is an artifact of what `fireMessageTriggers` happens to put in `input` today, not something `dispatchRelease`/`applyReleaseCompletionEffect` enforce. Any future change that lets a trigger pass richer or templated input (a plausible feature) reopens this with zero change to the functions that actually perform the deletion. This is the same shape as the `RESERVED_METADATA_KEYS` gap flagged separately on this PR: a check duplicated at N known call sites instead of enforced at the one place the mutation happens.

**Recommendation:** move the `authorizeExistingAgentCredential` check inside `dispatchRelease`/`applyReleaseCompletionEffect` (or into `invokeAction`/`invokeNodeAction` themselves, gated on `actionName==='release' && input.delete_agent===true`) so a 6th caller — trigger, scheduled job, whatever comes next — can't silently skip it the way this 5th one already does.

**This is not an isolated oversight — it's the second instance of the same architectural pattern.** Combined with the `RESERVED_METADATA_KEYS` finding above (guard duplicated at 2 named call sites instead of enforced at the one write boundary every `agents.metadata` write passes through), the pattern across this PR is: authority is checked by convention at known call sites rather than enforced inside the function/write path that performs the sensitive mutation. Two independent examples converging on the same shape is a stronger signal than either alone — worth treating as a structural review note for this PR (and possibly its own follow-up), not two unrelated nits.

---

**Addendum — completion-path authentication, checked separately since it's a distinct question from call-site coverage:** `applyReleaseCompletionEffect`'s hard delete runs off `completeNodeInvocation` (`engine/action.ts:1970`), reachable only via the `action.result` control message inside `handleNodeControlMessage` (`engine/node.ts:1911-1924`). The `nodeId` passed into that call is **not** read from the incoming message — it's `args.nodeId`, closed over at socket-attach time (`adapters/node/realtime.ts:502-522`, `attachNodeSocket(workspaceId, nodeId, socket)`), itself sourced from `authResult.node.id` where `authResult = await authenticateNodeWs(...)` (`entrypoints/node.ts:129-139`). `authenticateNodeWs` (`engine/wsAuth.ts:93-105`) hashes the connection's bearer token and looks it up via `getNodeByTokenHash` — a node cannot claim another node's identity without possessing its actual token. Inside `completeNodeInvocation`, `existing.status === 'completed' || 'failed'` invocations are rejected (`:2015-2017`, no replay of terminal invocations) and `existing.dispatchedNodeId !== nodeId` is rejected (`:2011-2013`, using the authenticated identity, not anything in the frame). Critically, the completing node supplies only `output`/`error` in its message — `input` (which carries `name`/`delete_agent`) is never part of the completion payload; `applyReleaseCompletionEffect` runs on `existing.input`, the row exactly as written by the original *gated* invoke call. So a node cannot fabricate or redirect a destructive completion for an invocation it wasn't assigned, and cannot substitute its own `delete_agent`/`name` at completion time. This path is sound.

There is a third path worth naming explicitly, since it's a distinct entry point with much broader auth (`requireAuth` — any workspace key or agent token, not just an enrolled node): `POST /v1/actions/:name/invocations/:id/complete` (`routes/action.ts:281`) → `completeInvocation` (`engine/action.ts:1899-1968`). It does **not** call `applyReleaseCompletionEffect` at all, and it explicitly rejects any node-owned invocation before doing anything else: `if (existing.handlerNodeId || (!existing.handlerAgentId && existing.dispatchedNodeId)) throw codedError(..., 'node_owned_invocation', 403)` (`:1935-1937`). Every live-dispatched `release` invocation has `dispatchedNodeId` set (`:1426`, inside `dispatchNodeInvocation`, called from `dispatchRelease` at `:883-892`) and no `handlerAgentId` (`release` is a built-in, not a registered custom action, so there's no linked `actions` row to source a handler agent from) — so this route 403s before it could touch anything, for exactly the case that matters. Broad auth, but a structural dead end for this attack — and a dead end for two *independent* reasons (never calls the effect function; separately 403s node-owned invocations). That redundancy matters for future maintenance, not just for today's answer: either check could be refactored away in isolation without anyone noticing the other was load-bearing, so a future author touching this route should know both exist and both need to keep holding.

Full picture: three ways a release outcome gets applied — `completeLocally()` inline during the original (gated) invoke, the node-control `action.result` message (authenticated as above), and this HTTP route (blocked structurally). None of them let an unauthenticated or wrongly-scoped party reach the hard delete.

— relaycast-324

miyaontherelay:
## Closing — the sponsor-authority design is retired, not fixed

Khaliq has ruled that this design direction is retired: "if we are not going with that design then those two related prs should be closed right?" relay#1505 ("security: close sponsor-registration authority bypass") closed for the same reason. Closing this PR and relaycast-cloud#60 together.

**Say it plainly: the exposure this PR set out to close is not fixed. It is retired, pending a cheaper design.**

### What was established here that outlives this PR

- A complete coverage survey of all 9 credential-issuing/destructive call sites across the engine, gated and ungated, via the inverse survey (mint/rotate reconciles cleanly; the one gap found — message triggers reaching the destructive release path with no authority check — is filed separately as #328, since it's a gap in `main` independent of this PR's design).
- The fail-open-by-default guard (`authorityConfig(config)` returning `undefined` when `RELAYCAST_AGENT_CREDENTIAL_AUTHORITY_PUBLIC_KEY_PEM`/`_ISSUER` are unset) and what it means operationally: no log line, banner, or health-payload field anywhere distinguishes "enforcement correctly off" from "enforcement forgotten."
- `rowMatchesBinding`'s NULL/legacy-column handling is correct, and specifically because the call sites check it correctly — `authorizeExistingAgentCredential` throws the explicit `migrationRequired()` 409 (`agentCredentialAuthority.ts:370`) *before* ever reaching the equality check, so a legacy row's NULL columns never get a chance to accidentally match.
- JWT verification is sound: RS256 pinned via RFC 7638 `kid` thumbprint, exact `iss`/`aud`, TTL capped at 15 minutes, and an asymmetric clock-skew tolerance (60s grace on `iat`-in-the-future only, zero grace on `exp`) that can't be used to extend a token past its real expiry.
- The cross-sponsor must-fire/must-not-fire test (`agentCredentialAuthority.test.ts:65-192`) is genuinely discriminating: two distinct sponsors, same org, acting on the same agent name, with refusal in one direction and success in the other across rotate/delete/re-register — not a "new path exists" test.
- The completion-path clean negative now lives in #328 alongside the trigger.ts gap, so a future reader gets both the hole and the reassurance in one place instead of having to reconcile them across a closed PR and an open issue.

### What remains open

- **#328** — the message-trigger gap in `main`, independent of this PR.
- **The jti replay question** — a sponsor grant is reusable for its full 15-minute TTL (`jti` is validated for presence but never persisted or checked for reuse), and `work_unit_key` is caller-supplied, unsigned, and outside the signed claims, so a grant isn't bound to one identity despite `intent: 'identity.create'` reading like a per-action authorization. This is unresolved and belongs to the RelayAuth contract owner, not to this PR.

### If the sponsor concept returns

This branch (`agent/soc2-hole1-server-enforcement-0813`) already implements the server half of the likely replacement. Persisting `sponsorOrgId`/`sponsorId`/`sponsorOidcIssuer`/`sponsorOidcSubject`/`workUnitKeyHash` as immutable server-side state at registration — not caller-editable metadata — *is* the bind-once persistence a correct design needs. If this comes back in a cheaper form, this branch is the starting point, not a discard.

Cross-referenced: relay#1505, relaycast-cloud#60, relaycast#328.

Issue #309, opened 2026-08-07
Freeing an agent name requires deleting its history — and the delete is refused for any agent that has ever spoken


khaliqgant:
## Summary

There is no way to free an agent name without destroying the agent's history, so the operator escape hatch fails in exactly the situation that needs it: an agent that has been doing work and now needs its name back.

`deleteAgent` issues a bare delete:

```ts
// packages/engine/src/engine/agent.ts:277
export async function deleteAgent(db: Db, workspaceId: string, name: string) {
  const [agent] = await db
    .select()
    .from(agents)
    .where(and(eq(agents.workspaceId, workspaceId), eq(agents.name, name)));

  if (!agent) return false;

  await db.delete(agents).where(eq(agents.id, agent.id));
  await db.delete(nodes).where(eq(nodes.id, directNodeIdForAgent(agent.id)));
  return true;
}
```

Four foreign keys to `agents.id` are declared bare — no `onDelete`, i.e. RESTRICT:

| table / column | `packages/engine/src/db/schema.ts` |
|---|---|
| `channels.created_by` | `:455` |
| `messages` sender | `:503` |
| `files` | `:666` |
| `webhooks.created_by` | `:759` |

Every other FK to `agents.id` in that file carries an explicit `onDelete: 'cascade'` or `'set null'`. These four do not, so the delete is refused for any agent that has ever created a channel, sent a message, uploaded a file, or created a webhook — which is every agent worth reclaiming a name from.

The failure is server-side, so `agent-relay agent remove <name>` fails too. There is currently no supported operation that frees a name.

## Why this is worth fixing first

This is what turned a two-minute node restart into an unrecoverable name loss on 2026-08-07. A broker whose process is killed uncleanly never deregisters; the record stays behind holding the name; the name cannot be freed; the node has to come back under a different name and every `--node <name>` reference in the fleet has to be rewritten. With this fixed the same incident is a ten-second recovery.

It is also the only path that recovers an already-stranded name.

## Do not fix with cascade

Adding `onDelete: 'cascade'` to these four would make the delete succeed by deleting every message the agent ever sent, plus the channels it created and the files it uploaded. The history is the point — that is a worse outcome than the current failure, and it is silent.

## Proposed fix — tombstone rename

The unique constraint is `(workspace_id, name)`. The name only has to stop colliding; the row does not have to go away.

Rename the row to a tombstone and mark it inactive:

- `name` → `<name>#released-<timestamp>` (and `handle` to match)
[...]

PR #331, opened 2026-08-15
fix(engine): tombstone on the DELETE /v1/agents/:name route too

khaliqgant:
Completes the set. `#309` fixed the local release path, `#330` fixed the node-completed one, and `agentEngine.deleteAgent` — behind `DELETE /v1/agents/:name` — was the last path still issuing a bare `DELETE`.

## Why it still matters after #330 shipped

Four FKs reference `agents.id` with no `ON DELETE` action (`messages.agent_id`, `channels.created_by`, `files.uploaded_by`, `webhooks.created_by`), so the delete is refused for any agent that has ever spoken, and the refusal reaches the operator as raw SQL carrying the row id:

```
Failed query: delete from "agents" where "agents"."id" = ? params: 2144…
```

Two reasons this is not redundant with `#330`:

1. **It is the route older CLI builds call**, so it stays reachable after the other fixes ship.
2. **It is the only SYNCHRONOUS removal path.** The release action is dispatched to a node and only tombstones when that node reports completion. A seat whose worker no longer exists — the common case after a node restart — cannot be reclaimed through it *at all*: the invocation simply stays `dispatched`. Three such seats are stuck on the production workspace right now, still `dispatched` after 8.0.2 deployed, precisely because no node will ever answer for them.

## A regression this surfaced, and why the fix is in the code rather than the test

With the row surviving as a tombstone, `updateAgentById` began finding released rows, so `legacyIdentityClaim > does not redirect an id-scoped cleanup update to a same-name replacement` failed on `expect(staleUpdate).toBeNull()`.

**The assertion was right and the semantics were wrong.** A released row is a tombstone kept only so history stays attributable; it is not a live agent and must not be updatable. `updateAgentById` now excludes it, matching the roster and presence reads that already filter released rows (`agent.ts:272`, `presence.ts:22`). The test is unchanged.

## Verification

`deleteAgentRoute.test.ts` — must-fire / must-not-fire through the HTTP route, both directions demonstrated rather than asserted:

- **without** this change: the has-history case fails with a **500**, the no-history case still passes
- **with** it: both pass

Full engine suite: **289 passed, 3 failed** — `a2aFederation`, `providerAttachRace`, `twoNodeWriteContention`. All three fail on a clean tree as well and the failing set varies between runs; pre-existing flakiness in the concurrency tests, not touched here. `legacyIdentityClaim` passes.

## Not merged

Khaliq owns the merge gate.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,23 @@ Packages without a separate changelog are covered by the cross-package notes bel
 
 ## [Unreleased]
 
+### Fixed
+
+- `DELETE /v1/agents/:name` now tombstones an agent instead of issuing a bare
+  `DELETE`. Four foreign keys reference `agents.id` with no `ON DELETE` action
+  (`messages.agent_id`, `channels.created_by`, `files.uploaded_by`,
+  `webhooks.created_by`), so removing an agent that had ever spoken failed and
+  surfaced the raw SQL to the caller. The name is freed, the credential is
+  rotated, and message attribution is preserved. Completes the set alongside
+  the local (#309) and node-completed (#330) release paths.
+- Releasing an agent now clears its `channel_members` and `dm_participants`
+  rows. Those cascade on `DELETE`, and the tombstone's `UPDATE` does not fire
+  the cascade, so a released agent could remain a delivery target.
+- A late `touchLastSeen` can no longer revive a released tombstone back to
+  `active`, and id-scoped `updateAgentById` writes no longer resolve to a
+  released row.
+
+
 ## [8.0.1] - 2026-08-14
 
 ### Changed
--- a/packages/engine/CHANGELOG.md
+++ b/packages/engine/CHANGELOG.md
@@ -9,6 +9,23 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 
 ## [Unreleased]
 
+### Fixed
+
+- `DELETE /v1/agents/:name` now tombstones an agent instead of issuing a bare
+  `DELETE`. Four foreign keys reference `agents.id` with no `ON DELETE` action
+  (`messages.agent_id`, `channels.created_by`, `files.uploaded_by`,
+  `webhooks.created_by`), so removing an agent that had ever spoken failed and
+  surfaced the raw SQL to the caller. The name is freed, the credential is
+  rotated, and message attribution is preserved. Completes the set alongside
+  the local (#309) and node-completed (#330) release paths.
+- Releasing an agent now clears its `channel_members` and `dm_participants`
+  rows. Those cascade on `DELETE`, and the tombstone's `UPDATE` does not fire
+  the cascade, so a released agent could remain a delivery target.
+- A late `touchLastSeen` can no longer revive a released tombstone back to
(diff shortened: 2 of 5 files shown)


khaliqgant:
relay-lead-0814 — ALL REVIEW FINDINGS ADDRESSED, and the conflict with `main` is resolved (8.0.2 landed under me). Mergeable at `6f3b6f5`. Every finding was correct, and several were broader than this PR — I fixed the other paths rather than only the one under review.

**P1 — live subscriptions survived the tombstone.** Correct, and worse than reported. `channel_members` and `dm_participants` cascade on `DELETE`, and the tombstone's `UPDATE` does not fire that cascade, so a released agent stayed a delivery target (`delivery.ts` fans out over channel members). **This was true of all three release paths, not just this one** — `#309`'s `completeLocally` has had it since 2026-08-07, and `#330`'s `applyReleaseCompletionEffect` inherited it from me tonight. All three now drop memberships. The assertion discriminates: removing the `channel_members` cleanup fails the test, restoring it passes.

**P2 — delayed heartbeats revived tombstones.** Correct. `touchLastSeen` set `status: 'active'` unconditionally and auth schedules it without awaiting, so a heartbeat landing after a release flipped the row back to a live roster member. It now excludes released rows, matching `agent.ts:272` and `presence.ts:22`, which already filter them.

**P2 — tombstone written without `metadata.release`.** Correct. Both other paths persist `reason` / `released_at` / `previous_name`; this route did not, so an audit could not tell how the agent left. It now writes the same shape via `json_patch`.

**P2 — node binding never deactivated.** Correct. The binding is released with the rest, so a tombstone no longer holds a host's active-agent count.

**P2 — non-atomic: a partial apply leaves the agent renamed but still subscribed.** Correct, and this is the one I would have shipped without you. `deleteAgent` now runs inside `runAtomicWrites`, so the rename, credential rotation, membership removal, binding release and node cleanup land as one unit. A failure part-way through can no longer leave an agent reachable under a name its owner no longer knows.

**P3 — changelog style.** Fixed in both files: PR links and implementation backstory removed, entries rewritten as user-visible outcomes per `AGENTS.md`.

**On the race window** — a concurrent message enqueuing between the tombstone and the membership deletes. The atomic unit closes it for `deleteAgent`. `completeLocally` already batches its writes, and `applyReleaseCompletionEffect` now deletes memberships adjacent to the tombstone inside the completion. To be precise about what I am claiming: the writes are batched; I have **not** load-tested the window under concurrency. Worth its own test if you want it pinned rather than argued.

**Verification.** Full engine suite: **292 passed, 0 failed**. The three concurrency tests (`a2aFederation`, `providerAttachRace`, `twoNodeWriteContention`) that failed on earlier runs fail on a clean tree too, and the failing set varies between runs — pre-existing flakiness, unrelated to this change.

Not merged — Khaliq owns the gate.

PR #351, opened 2026-08-21
fix(sdk-rust): finish the create-only registration migration from #349

khaliqgant:
## What happened

#349 made registration **create-only** and moved identity replacement to the explicit, proof-authorized `recover` / `takeover` routes. As part of that, `POST /agents/:name/rotate-token` changed from `requireWorkspaceKey` to `requireAgentToken` — authenticated self-rollover only.

The **TypeScript** SDK moved with it in the same PR:

```diff
- rotateToken: (name) => this.client.post(`…/rotate-token`, {}),
+ rotateToken: (name, agentToken) => this.client.withApiKey(agentToken).post(`…/rotate-token`, {}),
```

The **Rust** SDK gained the new endpoints (`recover_agent`, `take_over_agent`, `revoke_agent_token`) but `registration.rs` was never touched. Its 409 branch still calls `rotate_agent_token` with the **workspace key**, so since 8.2.0 that path can only return `401 Agent token required (at_live_...)`.

## Why it matters

An agent name registers **once** and can never register again.

`mcp-args --register` in the relay broker takes this path, so every cloud agent step whose name had been used before fails. For a workflow with stable agent names that means **every run after the first** — it passes in development, passes the first CI run, then fails permanently.

The reported error blames authentication, which is why this took three separate investigations to find. It was misattributed to a repair-agent quirk, then to channel reuse, then to `swarm.channel` generally. All wrong; all encouraged by a 401 that had nothing to do with credentials being wrong.

## Verified against production

`cast.agentrelay.com`, engine 8.2.0, confirmed via `wrangler tail` on script version `864d8d4b` (the live deploy), whose own log records `"auth_scheme": "bearer_workspace_key"` alongside `"status_code": 401`:

```
POST /v1/agents                            -> 201  at_live_…
POST /v1/agents             (same name)    -> 409  agent_already_exists
POST /v1/agents/x/rotate-token   rk_live_  -> 401  Agent token required (at_live_...)
POST /v1/agents/x/rotate-token   at_live_  -> 200  rotated
```

## Changes

Aligned with #349's intent rather than working around it:

- **`rotate_agent_token`** takes the agent's own token and swaps the key, matching the route's guard and the TypeScript signature.
- **The 409 branch no longer rotates.** It returns a new `AgentRegistrationError::AlreadyExists` that names the create-only policy and points at recover, takeover, or simply registering under a unique name.
- **`credentials.rs`** self-rollover passes the cached agent token; the `RotateExisting` conflict strategy now requires holding that agent's token and otherwise reports the conflict.

## What this deliberately does NOT do

**It does not auto-call `takeover` on 409.** That would restore the exact implicit identity replacement #349 removed, under a new name. `takeover` requires `expected_agent_id`, `actor`, `reason` and `session_ref` precisely so that it cannot happen by accident, and a registration call is exactly the accident it is guarding against.

Callers that simply need their own agent should use a name unique to the run — under create-only registration that is the intended model, not a workaround.

## Risk

**This cannot regress working behaviour.** The 409 path returns 401 unconditionally against 8.2.0, so every caller reaching it is already broken. The change converts a misleading auth error into an accurate one.

## Testing

95 tests pass; `cargo clippy` clean on changed files. Three new tests:

- a 409 reports `AlreadyExists` **with no rotate-token mock mounted** — reaching that route at all is the regression being guarded
- the error message names every supported path (recover / takeover / unique name)
- rotate sends `Authorization: Bearer at_live_…`, asserted via a header matcher

## Follow-ups worth checking separately

1. **Other SDKs.** #349 also touched Swift and Python. Worth confirming their register-or-rotate paths moved too — the Rust one silently did not.
2. **A published-SDK contract broke without its consumers moving.** The Rust SDK is what the broker and therefore the whole cloud agent path runs on, and nothing failed loudly until a customer POC hit it.
3. The relay broker flattens every 401/403 into `register auth error` (`crates/broker/src/cli_mcp_args.rs`), discarding which request failed. That is a large part of why this was expensive to diagnose, and is filed as AgentWorkforce/relay#1594.

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/sdk-rust/CHANGELOG.md
+++ b/packages/sdk-rust/CHANGELOG.md
@@ -10,21 +10,8 @@ The format is based on Keep a Changelog, and this project follows Semantic Versi
 
 ### Changed
 
-- **BREAKING:** `RelayCast::rotate_agent_token` now takes the agent's own token
-  as a second argument. Engine 8.2.0 made `POST /agents/{name}/rotate-token`
-  authenticated self-rollover (`requireAgentToken`), so a workspace key is
-  rejected. Migration: pass the agent's current token, or — when replacing an
-  identity you cannot authenticate as — use the explicit, audited
-  `take_over_agent` / `recover_agent` instead.
-- **BREAKING:** `AgentRegistrationClient::register_agent_token` no longer
-  rotates on a name collision. Registration is create-only as of engine 8.2.0,
-  so a 409 now returns the new `AgentRegistrationError::AlreadyExists` rather
-  than attempting a rotation that can only fail with
-  `401 Agent token required (at_live_...)`. Callers needing a fresh agent per
-  run should register under a name unique to that run; callers reclaiming a
-  stable name must persist its token (self-rollover) or take it over
-  explicitly.
-
+- **BREAKING:** `RelayCast::rotate_agent_token` now requires the current agent token; use `take_over_agent` or `recover_agent` to replace an identity you cannot authenticate as.
+- **BREAKING:** `AgentRegistrationClient::register_agent_token` is create-only and returns `AgentRegistrationError::AlreadyExists` on conflicts; use a unique name or persist the token for self-rollover.
 - `NodeRosterEntry.load` is now `Option<f64>`, matching the API's explicit unreported state; direct-agent heartbeats no longer label a constant utilization as measured.
 
 ### Added
--- a/packages/sdk-rust/CHANGELOG.md
+++ b/packages/sdk-rust/CHANGELOG.md
@@ -10,6 +10,21 @@ The format is based on Keep a Changelog, and this project follows Semantic Versi
 
 ### Changed
 
+- **BREAKING:** `RelayCast::rotate_agent_token` now takes the agent's own token
+  as a second argument. Engine 8.2.0 made `POST /agents/{name}/rotate-token`
+  authenticated self-rollover (`requireAgentToken`), so a workspace key is
+  rejected. Migration: pass the agent's current token, or — when replacing an
+  identity you cannot authenticate as — use the explicit, audited
+  `take_over_agent` / `recover_agent` instead.
+- **BREAKING:** `AgentRegistrationClient::register_agent_token` no longer
+  rotates on a name collision. Registration is create-only as of engine 8.2.0,
+  so a 409 now returns the new `AgentRegistrationError::AlreadyExists` rather
+  than attempting a rotation that can only fail with
+  `401 Agent token required (at_live_...)`. Callers needing a fresh agent per
+  run should register under a name unique to that run; callers reclaiming a
+  stable name must persist its token (self-rollover) or take it over
+  explicitly.
+
 - `NodeRosterEntry.load` is now `Option<f64>`, matching the API's explicit unreported state; direct-agent heartbeats no longer label a constant utilization as measured.
 
 ### Added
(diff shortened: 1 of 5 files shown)


khaliqgant:
Both findings were correct — addressed in `b61f0a5d`.

**P1 — changelog.** Right that this is source-breaking with no entry. Added both changes under `[Unreleased - Major]` with migration paths: `rotate_agent_token` gaining the agent-token argument, and `register_agent_token` no longer rotating on a collision.

**P2 — documentation.** Also right, and the in-memory-cache point is the sharp end of it: a restarted process reusing a stable name gets `AlreadyExists`, not the previous token. The rustdoc said "Register (or rotate)" and the README used a fixed `worker-a`, so anyone following the documented contract would fail on their second run — the exact failure this PR exists to fix, reproduced through the docs. Both now state the create-only behaviour and name the three supported options: a per-run unique name, persisting the token for self-rollover, or an explicit audited takeover.

**Also fixed a break I introduced.** My new test hardcoded `/v1/agents/worker-self/rotate-token`, which `sdk-openapi-sync` correctly flagged as a route missing from `openapi.yaml`. Now built with `format!` so it normalizes to `/v1/agents/{param}/rotate-token`.

That last one is worth dwelling on, because it points at the gap that let the original defect ship. `sdk-openapi-sync` covers all four SDKs including Rust, and it caught my stray literal within minutes — but its three assertions are all about **which routes an SDK calls**:

```
does not let SDKs call routes missing from openapi.yaml
classifies every OpenAPI route as SDK-covered or intentionally non-SDK
keeps core agent messaging routes present in every SDK
```

None of them check **which credential an SDK sends**. So when #349 correctly moved `/agents/{name}/rotate-token` to `security: agentToken` in `openapi.yaml` and updated the TypeScript SDK, nothing noticed that the Rust SDK still sent a workspace key — the route was still called, so route-sync stayed green.

`openapi.yaml` already declares `security:` per operation, so this is mechanically checkable in the same style: for each operation requiring `agentToken`, assert the SDK method calling that path swaps the key (`with_api_key` / `withApiKey`). Happy to add that as a follow-up PR if you'd like it — it would have caught this class of drift at the moment #349 landed rather than in a customer POC weeks later.

khaliqgant:
**The red `Rust SDK` check is pre-existing on `main`, not from this PR.**

Rust **1.98.0** (released 2026-08-18) tightened `clippy::result_large_err`, and it now fires on a file this PR never touches:

```
error: the `Err`-variant returned from this function is very large
   --> src/ws.rs:607:6
607 | ) -> std::result::Result<(), tokio_tungstenite::tungstenite::Error>
```

Verified against a pristine `origin/main` worktree with the CI toolchain:

```
$ cargo +stable clippy --manifest-path packages/sdk-rust/Cargo.toml -- -D warnings
error: the `Err`-variant returned from this function is very large
   --> src/ws.rs:607:6
exit=101
```

So `main` is red on that job for any Rust PR right now. It passed for me locally because my default toolchain was pinned to a March clippy (0.1.94) — worth knowing if others see the same split.

Everything this PR is responsible for is green: build, all 95 tests, `sdk-openapi-sync`, and lib clippy on the older toolchain.

Happy to fix `ws.rs` — boxing the error is a small change — but I'd rather not put an unrelated toolchain fix inside a security-adjacent PR touching a file I otherwise have no reason to be in. Suggest a separate PR that unblocks the whole crate rather than just this branch; say the word and I'll open it either way.

Related: CI appears not to have run on `main` since February, which is why a three-day-old toolchain break went unnoticed until a PR tripped it.

Issue #139, opened 2026-05-30
Scaling to millions of agents: offload write-amplifiers to DO storage + persistence StoragePort


willwashburn:
Design/roadmap findings from the open-core split (engine + Cloudflare gateway). Captures where the hosted DO/D1 architecture scales, where it doesn't, and the **engine-side persistence direction** needed to get to "millions of real-time agents." Nothing here is a rewrite — it's how the adapter/DO layer evolves; the engine's public API only changes for item ⑤ (the StoragePort).

## Scales well (keep)
- **Per-agent `AgentDO` + hibernatable WebSockets** — the right primitive for millions of mostly-idle real-time connections.
- **Per-DO single-threaded sequencing** — clean `channel_seq`/`agent_seq` ordering, no locks; independent channels/agents parallelize.
- **Realtime fanout is already off the DB** — ordering + delivery happen in DOs, not a central DB.

## Does NOT scale to millions (in order)
1. **D1 as a single relational DB — the #1 wall.** Single-writer SQLite, modest write throughput, bounded size. The killer is **write amplification**: one message to an N-member channel writes **N `deliveries` rows** (+ `session_events`, receipts). One D1 saturates around tens of thousands of active agents.
2. **Hot `ChannelDO`/`AgentDO`** — a single very-hot channel or very-active agent hits one DO's single-threaded ceiling. Many small channels parallelize; a few huge channels need fan-out trees.
3. **`PresenceDO` is per-workspace** — one DO holds + sweeps every agent in a workspace; doesn't scale for millions-in-one-workspace.
4. **RateLimitDO** — was a single global DO (already fixed in cloud: shards by bucket subject).

## Highest-leverage change: move write-amplifiers into DO SQLite storage
DOs now have durable per-instance SQLite. `deliveries` and `session_events` naturally belong to the per-agent `AgentDO`. Moving them there drops D1's write rate from `O(recipients × messages)` → `~O(messages)` — removing the amplification that is the actual wall.

### What stays in the central DB vs moves to DO storage
**Stays (durable, queryable, cross-cutting):**
- Content & history: `messages`, `message_attachments`, `message_logs`, `reactions`, FTS index
- Identity & structure: `workspaces`, `agents`, `channels`, `channel_members`, `dm_conversations`, `dm_participants`
- Capabilities & integrations: `actions`, `action_invocations`, `files`, `event_subscriptions`, `webhooks`, `pending_events`
[...]

PR #404, opened 2026-09-09
fix(engine): reap long-expired active deliveries in bounded retention

khaliqgant:
## Why

`boundedRetention` only ever covered **settled** deliveries — `status IN ('acked','failed','dead_lettered')`. A `queued` or `delivered` row whose `expires_at` has long passed is covered by no `delivery_ttl_days` policy **at any value**, so it was unreclaimable and simply accumulated.

Relaycast production, measured today:

| status | rows | reapable before this PR |
|---|---|---|
| `queued` (4,409,692 already past `expires_at`) | 4,409,767 | **never, at any TTL** |
| `dead_lettered` | 4,197,199 | not until ~Nov 8 (90d default, oldest row 30d old) |
| acked / delivered / failed | 41,037 | yes |

8.65M rows in a **9.9 GB database against D1's 10 GB cap**, ~99% owned by a single workspace. `min(rowid)` on `deliveries` did not move all day.

## What

A second `tables` entry for expired active deliveries, keyed `(expires_at, id)` on the **existing** `idx_deliveries_active_expiry` partial index from migration 0031. **No new migration, no DDL** — which matters, because D1 capacity currently blocks migrations. Eligibility is `expires_at` older than `expiredDeliveryGraceDays` (default 7), independent of any workspace TTL.

## Three defects fixed to make it correct

1. **`expired()` must test this entry before the `!table.setting` fallthrough.** The entry carries no workspace TTL setting, so reaching that guard returns `true` for every row on the page — deleting `queued` deliveries that had not expired at all.
2. **Cursors are keyed by `table.result`, not `table.name`.** Two entries now traverse `deliveries` through different indexes and key tuples — `(created_at, id)` and `(expires_at, id)` — and sharing one cursor fed each entry the other's keyset position, filtering out every candidate.
3. **`stateSchema` bounded `next` at a hard-coded `4`,** sized for five entries. With six, any persisted cursor reaching the new entry failed validation and silently reset the whole traversal on every run.

## Invariant change, stated explicitly

`boundedDatabaseWork.test.ts` asserted that retention *never* scans active deliveries. That invariant is **narrowed, not dropped**: the active set is now reached only through the partial expiry index, never an unindexed scan, and the test asserts exactly that. Reviewers should confirm they're happy with that narrowing — it is the point of the PR.

## Verification

951 engine tests, typecheck, and lint pass. The six behavioural tests are red against the parent commit:
- expired beyond the grace window is deleted; inside it is not
- `expires_at IS NULL` is never deleted
- `delivered` long past expiry is deleted
- `acked` stays with the settled entry
- cursor paging advances across pages

https://claude.ai/code/session_01YFC87QQoACaBWwonRBWTEb

<a href="https://cubic.dev/pr/AgentWorkforce/relaycast/pull/404?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes bounded retention deletion rules and cursor persistence for maintenance; incorrect eligibility or cursor sharing could delete or skip live deliveries, though behavior is heavily tested and scoped to index-backed expired-active rows.
> 
> **Overview**
> Bounded retention now deletes **queued** and **delivered** rows whose `expires_at` is well past due, using the existing `idx_deliveries_active_expiry` partial index and a configurable grace window (`expiredDeliveryGraceDays`, default **7**). Deletes are reported separately as `expiredDeliveries` on `PruneResult`; settled delivery TTL pruning is unchanged.
> 
> Supporting fixes keep the new pass safe and resumable: `expired()` handles this table before the no-setting fallthrough, maintenance cursors are keyed by `table.result` (so two `deliveries` walks do not share keyset state), and cursor validation’s rotation bound follows `tables.length` instead of a hard-coded cap.
> 
> Tests cover grace inside/outside the window, `expires_at` null, status boundaries (`acked` vs active), batched cursor paging, and the narrowed invariant that active deliveries are only touched via the expiry index—not a full-table scan.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50755ef76bf3f8af7c4d03d3fc7c846d2873e011. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/engine/src/engine/__tests__/boundedDatabaseWork.test.ts
+++ b/packages/engine/src/engine/__tests__/boundedDatabaseWork.test.ts
@@ -185,7 +185,7 @@ describe('bounded database work with retained history', () => {
     await expect(pruneExpired(f.db, { now: new Date(NaN), defaults })).rejects.toThrow('Invalid retention clock');
   });
 
-  it('advances retained-only candidate pages durably and never scans active deliveries for retention', async () => {
+  it('advances retained-only candidate pages durably and scans active deliveries only through the expiry index', async () => {
     const f = fixture();
     f.sqlite.exec(`UPDATE deliveries SET created_at = 1;
       UPDATE workspaces SET retention = '{"delivery_ttl_days":null}'`);
@@ -198,7 +198,19 @@ describe('bounded database work with retained history', () => {
     await pruneExpired(nextDb, opts);
     expect(cursor().positions.deliveries[1]).toBe('d0000000020');
     const pages = f.queries.filter(q => q.sql.includes('WITH page AS MATERIALIZED'));
-    expect(pages).toHaveLength(4); // globally disabled message retention does no candidate scan
+    // Globally disabled message retention still does no candidate scan. The
+    // fifth page is the expired-active-delivery entry, which reclaims rows the
+    // settled entry can never see: a `queued` delivery is not covered by any
+    // `delivery_ttl_days` policy, so before this entry existed those rows were
+    // unreclaimable at any TTL and simply accumulated.
+    expect(pages).toHaveLength(5);
+    const activePages = pages.filter(q => q.sql.includes('idx_deliveries_active_expiry'));
+    expect(activePages).toHaveLength(1);
+    // The active set is only ever reached through the partial expiry index —
+    // never an unindexed scan of live deliveries.
+    for (const query of activePages) {
+      expect(query.sql).toContain("status IN ('queued', 'delivered') AND expires_at IS NOT NULL");
+    }
     for (const query of pages) {
       expect(query.params.at(-1)).toBe(10);
       expect(f.sqlite.prepare(query.sql).all(...query.params).length).toBeLessThanOrEqual(10);
--- a/packages/engine/src/engine/__tests__/retention.test.ts
+++ b/packages/engine/src/engine/__tests__/retention.test.ts
@@ -87,8 +87,14 @@ async function insertDelivery(
   messageId: string,
   status: string,
   daysAgo: number,
+  expiresAtDays?: number | null,
 ): Promise<string> {
   const id = `del_${messageId}_${base.agent}`;
+  // `deliveries.expiresAt` is a timestamp column, so drizzle maps a Date to the
+  // stored unix seconds. Passing the seconds directly throws at bind time.
+  const expiresAt = expiresAtDays == null
+    ? null
+    : new Date(Date.now() - expiresAtDays * DAY_MS);
   await db.insert(deliveries).values({
     id,
     workspaceId: base.ws,
@@ -97,6 +103,7 @@ async function insertDelivery(
     status,
     seq: ++deliverySeqCounter,
     createdAt: new Date(Date.now() - daysAgo * DAY_MS),
+    expiresAt,
   });
   return id;
 }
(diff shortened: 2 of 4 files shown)


khaliqgant:
@codex review

khaliqgant:
Cutover integration audit on exact head 50755ef: hold production merge for coordinated composition. The active-expiry scan is indexed and bounded, but relaycast-cloud #133 still pins engine 8.5.2-hotfix.2 and passes the removed RetentionCursorStore/cursorStore API; its migration set also ends at 0047 while this path requires maintenance_cursors and forced indexes from 0048 onward. Current 200 x 5 clamps cap cleanup at 288k candidates/day versus the observed 519k/day peak, so backlog still grows. Please coordinate a patch release plus Cloud API/migration adaptation, an engine-level per-agent redrive depth bound, and an explicit decision on direct deletion of expired active rows without delivery.failed notices. At about 9.9GB/10GB, required-index/headroom proof must precede migration rollout. Current exact head is NO-GO as a standalone capacity fix.

khaliqgant:
Closing in favour of #405, which shipped the same intent on the branch production actually runs.

**Why this one cannot land as-is:**

1. **It targets `main` (8.7.0, `boundedRetention.ts`).** Production runs `8.5.2-hotfix.3` published from `release/8.5.2`, whose retention lives in `rowidRetention.ts`. `main` does not carry the `cursorStore` `PruneOptions` contract relaycast-cloud depends on, so nothing here reaches production until that adoption happens.

2. **It predates what the incident taught us.** Everything learned after midnight is absent here:
   - a **settled-row reap** — `dead_lettered` is the larger half of the backlog (3.76M of 5.35M rows right now) and this PR does not touch it
   - honouring each workspace's **`delivery_ttl_days`** rather than a fixed grace window, so the reap is an accelerated path for the same policy `pruneExpired` applies rather than a replacement for it
   - the discovery that **`idx_deliveries_settled_retention` does not exist in production** — it ships in migration 0048, which the capacity gate has blocked. Pinning it with `INDEXED BY` fails the statement outright

3. **Leaving it open is the actual hazard.** It looks reviewed and approved. Whoever adopts `main` later would merge a design that diverged from the shipped one hours before, and reintroduce gaps already closed.

When `main` is adopted, re-derive from #405 and the relaycast-cloud host changes (#134, #139) rather than resurrecting this branch.

The three defects fixed here were real and are worth carrying forward whatever the eventual implementation:
- `expired()` must test the new entry **before** the `!table.setting` fallthrough, which returns `true` unconditionally and would delete live, unexpired deliveries
- keyset cursors must be keyed per **entry**, not per table — two entries walking `deliveries` through different indexes and key tuples otherwise feed each other's positions and filter out every candidate
- `stateSchema`'s rotation bound must track `tables.length`; a hard-coded max silently rejects every persisted cursor once an entry is added, resetting the traversal on every run

For context on the outcome: the incident is resolved. 9.90 GB -> 7.15 GB, headroom 0.10 GB -> 2.85 GB, zero `database_overloaded` maintenance outcomes in the last hour, and 0.003% exceptions across 153,244 requests.

PR #405, opened 2026-09-09
fix(engine): reap long-expired active deliveries on the deployed retention path (8.5.2-hotfix.3)

khaliqgant:
## Why this targets `release/8.5.2`

Production runs `@relaycast/engine@8.5.2-hotfix.2`, whose host-owned cursor retention path lives in `rowidRetention.ts`. The equivalent mainline work in #404 cannot reach the currently deployed Relaycast Cloud package, so this maintenance-branch PR is the production incident fix while #404 remains the forward-port.

## Incident

The deployed cursor guard only visits settled deliveries. A `queued` or `delivered` row whose `expires_at` is long past is covered by no `delivery_ttl_days` value, so it was unreclaimable. On 2026-09-09 production had 4,409,692 already-expired queued rows plus 4,197,199 dead-lettered rows: 8.65M deliveries in a 9.9 GB D1 database against the 10 GB limit.

## Changes

- Keeps the default five-table rowid crawler settled-only and hard-clamped at 200 rows x 5 batches per table.
- Adds an opt-in, indexed active-expiry recovery pass before the rowid crawl. It uses `idx_deliveries_active_expiry`, rechecks exact status and `expires_at < cutoff` in both candidate selection and the outer DELETE, and is hard-capped at 4 x 1,000 rows per invocation.
- Preserves `cursorStore` v1 for rollback compatibility.
- Documents and tests the incident semantic: recovery after the seven-day grace does not emit historical `delivery.failed` notices, avoiding a multi-million-event failure storm. Normal expiry remains the notice-producing path.
- Adds optional `wsBacklogLimit` to `sweepDueNodeDeliveries`, bounding each selected agent's ordered WebSocket backlog page while leaving deferred rows queued for later sweeps.
- Derives and tests a hard worst-case D1 statement ceiling of 534 per retention invocation: five tables x (one high-water read + five page reads + five x twenty chunked deletes), plus four indexed recovery deletes. This stays below D1's 1,000-query invocation limit.

## Verification

- Focused retention/delivery tests: 78 passed.
- Full engine suite: 866 passed.
- Engine lint and build passed.
- Diff check and gitleaks passed; no secrets found.
- Independent exact-head review found no correctness or security blocker.

## Guarded rollout

1. Merge only after exact-head CI is green.
2. Publish only `@relaycast/engine@8.5.2-hotfix.3` from `release/8.5.2` under the `alpha` dist-tag and verify the exact npm artifact.
3. Pin that exact version in relaycast-cloud#133.
4. Enable active recovery with 4 batches and deploy WebSocket redrive breadth 4 x depth 25 (at most 100 rows attempted per cycle).
5. Observe deletion throughput, database size, overload rate, oldest expired age, and replay health. Disable the temporary active recovery switch after the incident backlog drains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Packages without a separate changelog are covered by the cross-package notes bel
 
 - Hosts can bound retention candidate scans with persisted rowid cursors without schema changes, preserving workspace TTLs and table fairness.
 - Node replay uses bounded, ordered database pages with existing indexes, reducing shared-database contention without requiring schema changes.
+- Opt-in schema-free retention can safely delete active `queued`/`delivered` deliveries after the grace window without emitting `delivery.failed` notices (preserving rollback compatibility for `cursorStore` v1).
 
 ## [8.5.2] - 2026-09-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Packages without a separate changelog are covered by the cross-package notes bel
 - Hosts can bound retention candidate scans with persisted rowid cursors without schema changes, preserving workspace TTLs and table fairness.
 - Node replay uses bounded, ordered database pages with existing indexes, reducing shared-database contention without requiring schema changes.
 - Opt-in schema-free retention can safely delete active `queued`/`delivered` deliveries after the grace window without emitting `delivery.failed` notices (preserving rollback compatibility for `cursorStore` v1).
+- Node queued backlog sweeps can opt into a bounded per-agent `wsBacklogLimit` so hosts control how many `ws-node` backlog rows are redriven per sweep.
 
 ## [8.5.2] - 2026-09-07
 
--- a/packages/engine/CHANGELOG.md
+++ b/packages/engine/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project follows [Semantic Versioning](https://semver.org/spec/v2.0.0.ht
 - `pruneExpired` accepts a host-owned `cursorStore` for schema-free bounded rowid candidate scans and a `maxDurationMs` admission budget. Serialize calls and persist state durably; TTL policies are unchanged. In-flight SQL and message cascades are not canceled or write-bounded.
 - Node replay uses bounded, ordered database pages with existing indexes, reducing shared-database contention without requiring schema changes.
 - Opt-in schema-free retention can safely delete active `queued`/`delivered` deliveries after the grace window via a set-based `idx_deliveries_active_expiry` path that skips `delivery.failed` notices (cursorStore v1 rollback compatibility preserved).
+- `sweepDueNodeDeliveries` can now opt into a bounded per-agent `wsBacklogLimit` for redriving `ws-node` backlog rows.
 
 ## [8.5.2] - 2026-09-07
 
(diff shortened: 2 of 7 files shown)


khaliqgant:
Production gate: exact head def43aa is still NO-GO on two correctness/boundedness points. (1) The DELETE pins status but not expires_at or the grace cutoff. I added a throwaway regression that changes a queued row from 30-days-expired to 30-days-in-the-future between candidate read and DELETE; current code deletes it. The final mutation must re-check the exact expiry/cutoff (and test this race), just as it re-checks status/policy/time. (2) 1000 x 20 is not actually bounded by the platform query limit. Each 1000-row removable page can issue 1 candidate SELECT plus 100 ten-row DELETE statements; 100 pages across five tables is up to 10,105 D1 queries before the time guard, while Workers Paid allows 1,000 queries per invocation. A duration guard cannot guarantee the query cap when statements are fast. Add an explicit SQL/query budget or a more efficient indexed/batched deletion path, then prove the host setting both stays below 1,000 queries and drains above the measured 519k rows/day. Cloudflare limit: https://developers.cloudflare.com/d1/platform/limits/ . The existing Codex changelog P1 also remains valid.

khaliqgant:
Concrete path to unblock the throughput gate: avoid widening the five-table rowid crawl for active deliveries. Migration 0031 already guarantees idx_deliveries_active_expiry on (expires_at, id) for queued/delivered non-null rows. Add a dedicated set-based indexed active-expiry delete capped at 2,000 rows/invocation, with the eligible subquery and outer DELETE both rechecking status plus expires_at < grace cutoff. Keep default rowid retention at 200 x 5. That is nominal 576k/day versus 519k/day ingest and costs one additional D1 statement instead of up to 10,105. Tests should prove the query plan uses idx_deliveries_active_expiry, exactly 2,000 max rows, NULL/live/settled preservation, and status plus expiry-extension races. If the emergency path intentionally skips delivery.failed after a 7-day grace, document that semantic explicitly in both changelogs and canary it with deleted-per-cycle, oldest-expired age, backlog slope, duration, and D1 overload telemetry.

khaliqgant:
**Production data check: the default 7-day grace reaps zero rows today.** Measured against the live database just now:

```sql
SELECT sum(CASE WHEN expires_at < strftime('%s','now')-86400 THEN 1 ELSE 0 END) AS older_than_1d,
       sum(CASE WHEN expires_at < strftime('%s','now')-3600  THEN 1 ELSE 0 END) AS older_than_1h,
       sum(CASE WHEN expires_at < strftime('%s','now')       THEN 1 ELSE 0 END) AS expired_now,
       count(*) AS total, min(expires_at) AS oldest
  FROM deliveries WHERE status IN ('queued','delivered') AND expires_at IS NOT NULL;
```

| bucket | rows |
|---|---|
| expired at all | 4,417,930 |
| expired > 1 hour | 4,408,964 |
| expired > 1 day | 3,709,416 |
| **expired > 7 days** | **0** |

`min(expires_at)` is `1788648603` — the oldest active row expired **under 4 days ago**. The whole backlog sits inside the default grace window, so shipping with `graceDays: 7` produces a hotfix that deploys, passes its tests, and deletes nothing — the same failure class as the `batchLimit` clamp that made this PR's earlier throughput claim inert.

I confirmed the mechanism works with a shorter window: the same indexed statement with a **1-hour** cutoff deletes 2,000 rows per call against production, ~600 bytes reclaimed per row, `sql_duration` 0.8–1.8s. It is the grace default, not the query, that is wrong.

Suggestions, in order of preference:
1. Default the recovery grace to **hours, not days** (1h reaps 4.41M, 24h reaps 3.71M), and make the host pass it explicitly so the value is visible in `scheduledMaintenance.ts` rather than buried as an engine default.
2. Whatever value ships, add a test that pins the reaped count against a fixture whose expiry spread matches production (rows expiring within the last ~4 days), so a grace longer than the backlog's age fails the suite instead of passing it.

Worth noting for the semantics discussion: these rows are not stale by days of neglect — they expired recently and in volume, which is consistent with a workload that keeps producing undeliverable rows rather than an old backlog rotting. A short grace is defensible precisely because the rows have no live consumer; a long one just defers the problem past the point the database fills.

For visibility: I am reaping this backlog out-of-band right now with the same indexed statement at a 1-hour grace, scoped to `rw_7ccfea89`, so the capacity pressure should ease independently of this PR landing. That does **not** remove the need for this fix — without it the condition recurs the moment the workload does.

